PR #18, opened 2020-09-20
Docker container to check compatibility with PHP versions

raaaahman:
## Changes

This is a very basic approach to test the plugin with various PHP versions. 

## Usage

From `wp-notify` root folder:

```bash
docker build -t wp-notify .
```

[Update]
Optionaly, you can specify a PHP version. You should then use a specific tag so you can have as much different images as you want (default: latest patch of 7.4). Example:

```bash
docker build -t wp-notify:5.6 . --build-arg PHP_VERSION=5.6
```

Then install the dependencies with the composer provided inside the container (use the same name that you passed as argument of the `-t` option). Example:
 
```bash
docker run --rm --volume /path/to/wp-notify:/usr/src/app wp-notify:5.6 composer install
```

Then you can run the unit tests inside the container:

```bash
docker run --rm --volume /path/to/wp-notify:/usr/src/app wp-notify:5.6 vendor/bin/phpunit
```

_Notes_: 
- The `wp-notify` name can be set to whichever name you want for your docker image. Remember to use the same name for the `build` and `run` command.
[Update]
- You can try with different version of PHP, use the version number referenced on [this page](https://github.com/docker-library/docs/tree/master/php#supported-tags-and-respective-dockerfile-links), but don't put the `-` separated part like `-alpine` or `-cli`.
- PHP version 8 is available, however, the actual version of PHPUnit is not compatible with it.
- The files are actually stored on your filesystem, thanks to the path to the `wp-notify` project you provided to the `--volume` option (use absolute path).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,9 +1,13 @@
-FROM php:5.6-cli
+ARG PHP_VERSION=7.4
+FROM php:${PHP_VERSION}-cli
 
 WORKDIR /usr/src/app/
 
 COPY composer-install.sh /usr/src/app/
 
 RUN apt-get update
-RUN apt-get install wget
+RUN apt-get install wget -y
+RUN apt-get install zip -y
+RUN apt-get install git -y
+RUN apt autoremove -y
 RUN ./composer-install.sh
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,9 @@
+FROM php:5.6-cli
+
+WORKDIR /usr/src/app/
+
+COPY composer-install.sh /usr/src/app/
+
+RUN apt-get update
+RUN apt-get install wget
+RUN ./composer-install.sh
--- a/composer-install.sh
+++ b/composer-install.sh
@@ -1,0 +1,17 @@
+#!/bin/sh
+
+EXPECTED_CHECKSUM="$(wget -q -O - https://composer.github.io/installer.sig)"
+php -r "copy('https://getcomposer.org/installer', 'composer-setup.php');"
+ACTUAL_CHECKSUM="$(php -r "echo hash_file('sha384', 'composer-setup.php');")"
+
+if [ "$EXPECTED_CHECKSUM" != "$ACTUAL_CHECKSUM" ]
+then
+    >&2 echo 'ERROR: Invalid installer checksum'
+    rm composer-setup.php
+    exit 1
+fi
+
+php composer-setup.php --quiet
+RESULT=$?
+rm composer-setup.php
+exit $RESULT
--- a/composer-install.sh
+++ b/composer-install.sh
@@ -11,7 +11,7 @@ then
     exit 1
 fi
 
-php composer-setup.php --quiet
+php composer-setup.php --quiet --install-dir=/usr/local/bin/ --filename=composer
 RESULT=$?
 rm composer-setup.php
 exit $RESULT


szepeviktor:
It is called :) GitHub Actions.

And the answer to your next question :-D
https://github.com/nektos/act

Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM php:5.6-cli
+
+WORKDIR /usr/src/app/
+
+COPY composer-install.sh /usr/src/app/
+
+RUN apt-get update
+RUN apt-get install wget
+RUN ./composer-install.sh
```
youknowriad, line 13: Should we be using `.wp-env`/`@wordpress/env` instead of custom docker containers as you get the WP environment ready and you'd be able to use it for all tests and checks and even for development.

jonathanbossenger: @youknowriad I assume you are referring to this? https://make.wordpress.org/core/2020/03/03/wp-env-simple-local-environments-for-wordpress/

@raaaahman it might be worthwhile investigating this as an option. What I do like about it is it allows folks already familiar with the tool to help us develop and test

szepeviktor:
Please **do** use GitHub Actions  - they can be ran locally!! https://github.com/WordPress/wp-notify/pull/18#issuecomment-695773031

raaaahman:
> Please **do** use GitHub Actions - they can be ran locally!! [#18 (comment)](https://github.com/WordPress/wp-notify/pull/18#issuecomment-695773031)

I'm curious. 

What would be the benefits of local GitHub actions over a local environment?
Also, is there any interesting GitHub actions already set for Continuous Integration of WordPress plugins?

szepeviktor:
> What would be the benefits of local GitHub actions over a local environment?

- you don't have to invent the wheel
- GHA is relatively simple to configure
- you don't have to maintain 2 environments (CI and local)
- Gutenberg has it! https://github.com/WordPress/gutenberg/tree/master/.github/workflows

Dockerfile
raaaahman: Thank you @youknowriad , this environment looks promising! I made the minimum necessary to use it here: https://github.com/WordPress/wp-notify/pull/22

For now though, we don't need a full WordPress website to be loaded. But we will soon!

Do you have any idea about when PHP 8 will be available?

Dockerfile
```diff
@@ -0,0 +1,13 @@
...
+WORKDIR /usr/src/app/
+
+COPY composer-install.sh /usr/src/app/
+
+RUN apt-get update
+RUN apt-get install wget -y
+RUN apt-get install zip -y
+RUN apt-get install git -y
+RUN apt autoremove -y
+RUN ./composer-install.sh
```
szepeviktor, line 13:
```suggestion
RUN set -e \
    && apt-get update \
    && apt-get install -y wget zip git \
    && apt autoremove -y \
    && ./composer-install.sh
```

With every new `RUN` command the image grows as Docker creates a new aufs layer.

szepeviktor: And yes, 🐳 it is ugly.

szepeviktor (changes requested):

raaaahman:
Experiencing some technical difficulties at the moment:

Running `docker run -it --rm -v /path/to/wp-notify:/usr/src/app wp-notify:8.0 php /usr/local/bin/composer self-update --2 && php /usr/local/bin/composer update --with phpunit/phpunit:^9` returns:

 ```
Your requirements could not be resolved to an installable set of packages.

  Problem 1
    - phpunit/phpunit[9.3.9, ..., 9.5.x-dev] require php >=7.3 -> your php version (7.2.11) does not satisfy that requirement.
    - phpunit/phpunit[9.0.0, ..., 9.2.6] require php ^7.3 -> your php version (7.2.11) does not satisfy that requirement.
    - phpunit/phpunit[9.3.0, ..., 9.3.8] require php ^7.3 || ^8.0 -> your php version (7.2.11) does not satisfy that requirement.
    - Root composer.json requires phpunit/phpunit *, ^9 -> satisfiable by phpunit/phpunit[9.0.0, ..., 9.5.x-dev].
```

Even though running `docker run -it --rm -v /path/to/wp-notify:/usr/src/app wp-notify:8.0 php -v` returns:

 ```
PHP 8.0.0 (cli) (built: Dec  1 2020 03:14:26) ( NTS )
Copyright (c) The PHP Group
Zend Engine v4.0.0-dev, Copyright (c) Zend Technologies

```

While running `docker run -it --rm -v /path/to//wp-notify:/usr/src/app wp-notify:8.0 php -v && composer self-update --2 && php composer update --with phpunit/phpunit:^9` crashes with message:

```
[Composer\Downloader\FilesystemException]                                    
  Filesystem exception:                                                        
  Composer update failed: "/usr/local/bin/composer" could not be written.      
  rename(/home/sylvain/.cache/composer/composer-temp.phar,/usr/local/bin/comp  
  oser): Permission denied     
```
`

raaaahman:
Since `wp-env` now supports PHP 8.0, there is no need for a custom Dockerfile.

PR #19, opened 2020-09-20
Upgrade PHPUnit version

### LINKED ISSUE #12: Upgrade PHPUnit version used for Unit Tests
https://github.com/WordPress/wp-notify/blob/af086175c661586f691947886a519f8b550a6f5d/composer.json#L6

Because WordPress still needs to support PHP 5.6 environments, the maximum version we could use is 7.5.*.

The upgrade from 3 to 7 is expected to cause several errors in the tests code.

raaaahman:
## Update

Since the release of PHP 8, it is not possible to use a single PHPUnit version to test with PHP versions from 5.6 to 8.

We could draw inspiration from:

- `WordPress/wordpress-develop`: The WordPress Core team has adapted PHPUnit 7.5 to not break with PHP8.
- `woocommerce/woocommerce`: Uses quite the same trick, but goes further into creating its own fork of PHPUnit 7.5
- `Yoast/wordpress-seo`: Created a composer package to use different versions of PHPUnit given the PHP version tested.

## Changes

- Update the version of PHPUnit to 7.5.20.
- Integrates [JohnKary's Speedtrap Listener](https://github.com/johnkary/phpunit-speedtrap) instead of a custom one.

## Going further

- Actually, the versions of PHPUnit we can use is limited to 7.x as long as WordPress needs to support PHP 5.6. There is an [ongoing discussion about WordPress support for PHP on Trac](https://core.trac.wordpress.org/ticket/46149).
- To test with another PHP version than your current one, you can use the `experimental-docker` branch (but you will need to merge or cherry-pick this branch into your current working branch, remember to make a copy of it). See more info in #18

Fixes #12


jonathanbossenger:
@raaaahman I'm going to leave this open until we make a call on #18, as that decision might affect this one.

raaaahman:
> @raaaahman I'm going to leave this open until we make a call on #18, as that decision might affect this one.

I don't think so. The tests can run on local environment as well (they couldn't before).

jonathanbossenger:
> I don't think so. The tests can run on local environment as well (they couldn't before).

Aha, ok thanks. I will do my best to test this out in the coming week.

danbilauca (approved):

PR #20, opened 2020-10-10
Added simple plugin header, so that this can be installed as a plugin

jonathanbossenger:


wp-notify.php
```diff
@@ -1,5 +1,9 @@
 <?php
 
+/**
+ * Plugin Name: WP Notify
```
raaaahman, line 4: I like this name

raaaahman (commented):
During October 12, 2020 Office hours, two team names were proposed to fill the `Plugin Author` field:

1. `WP-Notify team`
2. `WordPress.org community`

My opinion was slightly geared toward `WordPress.org community` as it would feel more open for anyone to join any time.

Issue #16, opened 2020-09-14
Message data model discussion


folletto:
We need to collaborate across design and dev to clarify what's the data model we want to use. Having the data model helps inform design, in return design will check if the data model works for all our use cases.

The current draft is in code [here](https://github.com/WordPress/wp-notify/blob/develop/includes/Notification.php), translated in simpler terms, this is the i1 draft:

* `get_id` int
* `timestamp` timedate
* `sender` user_id
* `recipients` users? roles? mix?
* `message` text? html?

Let's review what we need to do, how much it matches the [requirements document](https://docs.google.com/document/d/1SoIaFqXkXiVzq5mizbZQafMfL36bD0WKj8iwYM823MI/edit), and how well it fits the use cases explored in the designs.

---

### Jump to Current Update
The discussion evolved, but no need to read through everything, just check the latest summary and proposal here:
https://github.com/WordPress/wp-notify/issues/16#issuecomment-724608463

folletto:
Reporting here a note from the [i1 design iteration](https://github.com/WordPress/wp-notify/issues/1#issuecomment-690501831): that design iteration has a lot going on in terms of data model, with optional fields an all. It might not be what we want, but here's the current needs to make that design possible:

* All messages have a `message` content
* Some `message` content is HTML
* Optional `title`
* Optional `action` (label + URL?)
* Optional `image` (direct SVG or URL)

Open for discussion.

folletto:
Reporting here also @daniloercoli's comment

—

Just my 2 cents about the data model from a mobile native perspective.

* If we go with HTML content in the body, i hope it will be restricted to few tags, otherwise it could be hard to represent the body properly in a native web view.
* Action - What the button should do on mobile native? i guess the easiest approach will be to open a web view, both in app or on system browser, and take the user to the proper HTML page within wp-admin.

raaaahman:
>     * All messages have a `message` content
> 
>     * Some `message` content is HTML
> 
>     * Optional `title`
> 
>     * Optional `action` (label + URL?)
> 
>     * Optional `image` (direct SVG or URL)

I like the idea to put image and action in the data model, rather than putting it in the HTML message. This could really ease the communication between frontend and backend IMO.

Although, I would say we should focus on simple messages without any options for v0.1, in order to get a proof of concept ready the sooner we can.

raaaahman:
[...]

PR #21, opened 2020-10-10
Adding initial database schema

jonathanbossenger:
Documenting the Notification table schema, based on the properties of WPNotify_BaseNotification

### DIFF
--- a/docs/database-schema.md
+++ b/docs/database-schema.md
@@ -5,9 +5,9 @@
 * sender `Sender of the notification.`
 * timestamp `Timestamp of when the notification was triggered.`
 * recipients `Collection of notification recipients.`
-* title `Notification message.` (optional)
-* message `Notification message.`
+* title_key `Notification title key.`
+* message_key `Notification message key.`  (optional)
 * action_link `Correctly formatted URL to an internal or external action.`
 * status enum (READ, UNREAD) `Notification status.`
 
-
+See the Translations document for details on title_key and message_key fields
--- a/docs/database-schema.md
+++ b/docs/database-schema.md
@@ -4,9 +4,9 @@
 * id (int) 11 auto_increment `ID of the notification.`
 * sender_key `Sender of the notification. As defined by the plugin or theme`
 * timestamp `Timestamp of when the notification was triggered.`
-* recipients `Collection of notification recipients.`
-* title_key `Notification title key.`
-* message_key `Notification message key.`  (optional)
+* recipient_key `Single notification recipient. As defined by the plugin or theme`
+* title_key `Notification title key. As defined by the plugin or theme`
+* message_key `Notification message key. As defined by the plugin or theme`  (optional)
 * action_link `Correctly formatted URL to an internal or external action.`
 * status enum (READ, UNREAD) `Notification status.`
 
--- a/docs/database-schema.md
+++ b/docs/database-schema.md
@@ -5,7 +5,9 @@
 * sender `Sender of the notification.`
 * timestamp `Timestamp of when the notification was triggered.`
 * recipients `Collection of notification recipients.`
+* title `Notification message.` (optional)
 * message `Notification message.`
+* action_link `Correctly formatted URL to an internal or external action.`
 * status enum (READ, UNREAD) `Notification status.`
 
 
--- a/docs/database-schema.md
+++ b/docs/database-schema.md
@@ -1,0 +1,11 @@
+# Database Schema
+
+### wp_notifications table
+* id (int) 11 auto_increment `ID of the notification.`
+* sender `Sender of the notification.`
+* timestamp `Timestamp of when the notification was triggered.`
+* recipients `Collection of notification recipients.`
+* message `Notification message.`
+* status enum (READ, UNREAD) `Notification status.`
+
+
--- a/docs/database-schema.md
+++ b/docs/database-schema.md
@@ -2,7 +2,7 @@
 
 ### wp_notifications table
 * id (int) 11 auto_increment `ID of the notification.`
-* sender `Sender of the notification.`
+* sender_key `Sender of the notification. As defined by the plugin or theme`
 * timestamp `Timestamp of when the notification was triggered.`
 * recipients `Collection of notification recipients.`
 * title_key `Notification title key.`
--- a/docs/translations.md
+++ b/docs/translations.md
@@ -1,0 +1,25 @@
+# Translations
+
+See also Database Schema
+
+To allow for internationalisation, the title and message text for a single notification is purposefully not stored in the database.
+
+Instead, a plugin or theme should store a key, which is linked to a translatable message in the plugin or theme's translation files.
+
+For example, given the following list of translatable strings in a plugin 
+
+```
+$notifications = array( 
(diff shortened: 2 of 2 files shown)


jonathanbossenger:
Final schema to be determined, see conversation on #16

docs/database-schema.md
```diff
@@ -0,0 +1,13 @@
...
+
+### wp_notifications table
+* id (int) 11 auto_increment `ID of the notification.`
+* sender `Sender of the notification.`
+* timestamp `Timestamp of when the notification was triggered.`
+* recipients `Collection of notification recipients.`
+* title `Notification message.` (optional)
+* message `Notification message.`
+* action_link `Correctly formatted URL to an internal or external action.`
+* status enum (READ, UNREAD) `Notification status.`
```
raaaahman, line 11: Beware that the `status` needs to be tied to each recipient. This leaves two possibilities:

1. The `recipients` field become `recipient` to only contain one recipient, and the notifications are duplicated for each recipient.
2. We create a relations table (let's name it `wp_sent_to`) which maps recipients to notification AND stores the `status` for each recipient-notification relation.

raaaahman (changes requested):

docs/database-schema.md
```diff
@@ -0,0 +1,13 @@
+# Database Schema
+
+### wp_notifications table
+* id (int) 11 auto_increment `ID of the notification.`
+* sender `Sender of the notification.`
+* timestamp `Timestamp of when the notification was triggered.`
+* recipients `Collection of notification recipients.`
+* title `Notification message.` (optional)
+* message `Notification message.`
```
swissspidy: How will title & message deal with internationalization?

If I send a notification to all users, it should be displayed in the user's language.

jonathanbossenger: @swissspidy thanks for raising this point.

If we consider the current admin_notices implementation, the plugin or theme author is responsible for making their strings translatable. So off the top of my head, I'm thinking, instead of storing the title and message as text, it should be some unique key, that's relevant to a list of notification messages that the plugin or theme stores in the code. 

For example, if I have some translatable strings in an array in my plugin somewhere

`
$notifications = array(
	'plugin_slug_new_podcast_title' => __('Some title', 'plugin-slug'),
	'plugin_slug_message' => __('Some message', 'plugin-slug'),
);
`
Then when I create the new notification I instead store the keys `new_podcast_title` and `new_podcast_message` instead of the actual text.

Then WP Notify, when it sends/displays the message, can lookup these keys to get the translatable text, and translate it as required.

Obviously, this would need to be thought about and documented properly, but hopefully, you get what I mean. Does this seem like a workable solution to you?

swissspidy: Yeah I think not storing the actual notification but just the key could work.

jonathanbossenger: Updated schema to reflect new keys, added Translations document with a brief description of how keys can be used.

docs/database-schema.md
```diff
@@ -0,0 +1,13 @@
+# Database Schema
+
+### wp_notifications table
+* id (int) 11 auto_increment `ID of the notification.`
+* sender `Sender of the notification.`
```
raaaahman: The sender is becoming a model of its own. See #41 

So it may be a `sender_key` as well.

docs/translations.md
```diff
@@ -0,0 +1,25 @@
...
+
+Instead, a plugin or theme should store a key, which is linked to a translatable message in the plugin or theme's translation files.
+
+For example, given the following list of translatable strings in a plugin 
+
+```
+$notifications = array( 
+    'plugin_slug_new_podcast_title' => __('Some title', 'plugin-slug'), 
+    'plugin_slug_message' => __('Some message', 'plugin-slug'), 
+);
```
raaaahman, line 15: WordPress i18n functions match translated strings with their source string (e.g.: 'Some title' ). We may be able to do the same to provide an easier API for plugin developers.

Note that this still holds the plugin responsible for returning the correct translation.

docs/database-schema.md
```diff
@@ -0,0 +1,13 @@
...
+* timestamp `Timestamp of when the notification was triggered.`
+* recipients `Collection of notification recipients.`
+* title_key `Notification title key.`
+* message_key `Notification message key.`  (optional)
```
raaaahman: Maybe this two fields should be coupled in the Message model.

docs/database-schema.md
jonathanbossenger: @raaaahman if we follow database normalization guidelines, we should probably create a separate table. However, I am concerned about the possible pushback from Core in creating a new table as part of (eventually) merging this feature. Are there any downsides with going with option 1?

docs/database-schema.md
jonathanbossenger: Left a comment on #41 to understand what the sender data will look like so that I can update this as needed.

docs/database-schema.md
jonathanbossenger: Is there an open issue discussing this model?

raaaahman: Not yet.

jonathanbossenger: @raaaahman if it's ok with you, let's merge this for now (as it's not set in stone) and we can iterate if things changed, based on the outcome of the message model discussion.

The main reason for this is that I can see the message data requirement will need to be both somehow stored by both the plugin and theme, but also needing to be dynamically generated as needed.

docs/translations.md
jonathanbossenger: @raaaahman forgive me for not understanding. I'm not sure if this means things should change here?

docs/database-schema.md
raaaahman: @jonathanbossenger I'd said that would be totally fine for the base implementation. The goal is for the design of the persistence layer to be modular, so crafty third party developers concerned by performance could replace it by whatever they fancy (Redis cache for example).

docs/translations.md
raaaahman: It would be easier for plugins to write just their string wrapped in a WordPress i18n functions, something like:

```php
$notification = $notification_factory->create( 
    array(
        'title' => __( 'Some title', 'plugin-domain' ),
        'message' => __( 'Some message', 'plugin-domain' )
    )
);
```

Not sure about how it'll work actually. :/

docs/database-schema.md
raaaahman: If we choose option 1 from [this comment](https://github.com/WordPress/wp-notify/pull/21/files#r503378289), we'd still need to modify the `recipients` to be a single `recipient` (or `recipient_key`), otherwise we set the `status` to READ/UNREAD for all the recipients at the same time. :)

docs/translations.md
jonathanbossenger: > Note that this still holds the plugin responsible for returning the correct translation.

I think this is the correct implementation, but it doesn't require any database schema changes to be possible.

jonathanbossenger:
@raaaahman I've updated the `recipients` to `recipient_key` as suggested, indicating that the data will be retrieved from a look-up on the key from the theme or plugin code/data. Please give this a once over, but otherwise, I think it's ready to merge.

raaaahman (approved):

raaaahman (approved):

PR #48, opened 2021-02-08
Feat(SenderModel): BaseSender model implemented

danbilauca:


includes/BaseNotification.php
```diff
@@ -185,10 +186,16 @@ public static function json_unserialize( $json ) {
...
 		$recipients       = new $recipients_class( current( $data ) );
+
 		next( $data );
 		$message_class = key( $data );
 		$message       = new $message_class( current( $data ) );
 
-		return new self( new WPNotify_BaseSender(), $recipients, $message );
+		next( $data );
+		$sender_class      = key( $data );
+		$sender_reflection = new ReflectionClass( $sender_class );
```
raaaahman, line 196: I'd rather avoid the cost of using PHP reflections inside the production code. What is the benefit of using `ReflectionClass` here?

tests/phpunit/tests/base-notification.php
```diff
@@ -61,5 +61,7 @@ public function test_it_can_be_instantiated_from_json() {
 			$testee->get_recipients() );
 		$this->assertInstanceOf( 'WPNotify_BaseMessage',
 			$testee->get_message() );
+		$this->assertInstanceOf( 'WPNotify_Sender',
```
raaaahman: Maybe you could check that the sender is a `WPNotify_BaseSender`, since this is the exact class you passed as property.

raaaahman (commented):
Good. Maybe you could also test and implement the unserialization of a `WPNotify_Sender`?

includes/BaseNotification.php
danbilauca: `$sender_class` is `WPNotify_BaseSender` string class name and we need to instantiate it dynamically with params. Check this json string:

```
{"WPNotify_RecipientCollection":[],"WPNotify_BaseMessage":"Message","WPNotify_BaseSender":{"name":"Name 1"}}
```
and observe the `WPNotify_BaseSender` object.

tests/phpunit/tests/base-notification.php
danbilauca: All over in `BaseNotification.php` file the `$sender` property is defined as being `WPNotify_Sender` and this was the reason why I've test it this way. Please check the definition of the `$sender` property and the return comment of `get_sender()` method.

Should I still need to test it as being `WPNotify_BaseSender` ?

includes/BaseNotification.php
raaaahman: Okay, now I see  it. :+1:

tests/phpunit/tests/base-notification.php
raaaahman: You are right, the `BaseNotification` class can indeed instantiate with any sub-class of `WPNotify_Sender`.

Although, it is the same with the `WPNotify_Message` class, and we are checking for a `WPNotify_BaseMessage` here: https://github.com/WordPress/wp-notify/blob/06d7be5918467b0c70c210976631de550b2aaba1/tests/phpunit/tests/base-notification.php#L62

I don't really know how the code will evolve at the moment, so I don't know what the better approach should be. Choose what makes the most sense to you. But I'd like to keep the code consistent, so could you please change it to either use child classes in every assertion, or parent classes in every assertion?

raaaahman (approved):

Issue #40, opened 2021-01-18
Pass the Sender as parameter to the Factory


raaaahman:
We should replace the following line to use a `Sender` instance from a `Factory` property instead:

https://github.com/WordPress/wp-notify/blob/3b5fca5d3b09cf820a9a2e5534113ea629d4342d/includes/Factory.php#L43

This would allow each plugin or theme to be identified as sender of notifications.

blocked by #41

PR #50, opened 2021-02-19
#40 Sender instance passed to the Factory

danbilauca:
I've chosen to implement a Sender Factory and send it along to Factory's constructor in order to have code consistency and have it processed/validated by Factory's logic.

### DIFF
--- a/includes/Factory.php
+++ b/includes/Factory.php
@@ -2,7 +2,7 @@
 
 class WPNotify_Factory {
 
-	const MESSAGE    = 'message';
+	const MESSAGE = 'message';
 	const RECIPIENTS = 'recipients';
 
 	/**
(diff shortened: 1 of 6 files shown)


includes/Factory.php
```diff
@@ -19,28 +19,45 @@ class WPNotify_Factory {
...
+	 *
+	 * @param $args
+	 *
+	 * @return WPNotify_BaseNotification
+	 */
 	public function create( $args ) {
-		list( $message_args, $recipients_args ) = $this->validate( $args );
 
-		$message = $this->message_factory->create( $message_args );
+		[ $message_args, $recipients_args, $sender_args ] = $this->validate( $args );
```
raaaahman: Short array syntax is not in the WordPress coding standards...

We should configure the PHP Code Sniffer to help us with these standards.

includes/senders/SenderFactory.php
```diff
@@ -0,0 +1,13 @@
...
+interface WPNotify_SenderFactory {
+
+	/**
+	 * Create a new instance of notification sender
+	 *
+	 * @param string $name
+	 *
+	 * @return WPNotify_Sender
+	 */
+	public function create( string $name ): WPNotify_Sender;
```
raaaahman: I'm not sure if we should use static type hinting or not.

What would be the benefits in your opinion?

tests/phpunit/tests/test-factory.php
```diff
@@ -0,0 +1,38 @@
...
+		                        ->getMock();
+
+		$message_factory->method( 'create' )->willReturn( $this->createMock( 'WPNotify_Message' ) );
+		$message_factory->method( 'accepts' )->willReturn( true );
+
+		$sender_factory = $this->getMockBuilder( 'WPNotify_SenderFactory' )
+		                       ->setMethods( array( 'create' ) )
+		                       ->getMock();
+
+		$sender_factory->method( 'create' )->willReturn( $this->createMock( 'WPNotify_Sender' ) );
```
raaaahman, line 21: Good. With all these mocks, you won't have to rewrite your test each time we change the code of any of these factories.

includes/Factory.php
danbilauca: I've fixed it by using array

includes/senders/SenderFactory.php
danbilauca: yep, fixed it by removing it

raaaahman (approved):

PR #51, opened 2021-03-03
Feature - image model #47

### LINKED ISSUE #47: Define the Icon / Image model
If we look at the designs for On-Hub notifications, there is two uses for images in the notifications:

- One is the icon that represents the **sender** (see [this comment]((https://github.com/WordPress/wp-notify/issues/41#issuecomment-773525936 )) in the Sender model discussion)
- The other is an image related to the message of the notification

Maybe both the Sender model and the Notification model would depend on a third model: the Image model?

Here are the data it should contain:

- **source**: string. An URL or data URL (see details below)
- **alternative_text**: string. Description of the image for accessibility purpose.

On the front-end, the `` tag should be able to handle all the use cases:

- An **external URL** will be resolved in a same way by its [src attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img#attr-src) whether the format of the image is PNG, JPEG or SVG.
- This `src` attribute can as well accept a [data URL](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/Data_URIs) to directly embed base64 encoded raster images or SVG inside the notification itself (saves an HTTP request, per notification!)

danbilauca:
## Goal

This PR implements the `WPNotify_BaseImage` model. 

Fixes #47

### DIFF
--- a/includes/image/BaseImage.php
+++ b/includes/image/BaseImage.php
@@ -35,7 +35,36 @@ class WPNotify_BaseImage implements WPNotify_Image, WPNotify_JsonUnserializable 
 		);
 	}
 
+	/**
+	 * Source of the image to be used on src attribute of <img/>
+	 * @return string
+	 */
+	public function get_source() {
+		return $this->source;
+	}
+
+	/**
+	 * Alternative text to be used on alt attribute of <img/>
+	 * @return string
+	 */
+	public function get_alt() {
+		return $this->alt;
+	}
+
+	/**
+	 * Instantiates a WPNotify_BaseImage based on a JSON string
+	 *
+	 * @param string $json
+	 *
+	 * @return WPNotify_JsonUnserializable
+	 */
 	public static function json_unserialize( $json ) {
-		// TODO: Implement json_unserialize() method.
+
+		$data = json_decode( $json, true );
+
+		$source = ! empty( $data['source'] ) ? $data['source'] : '';
+		$alt    = ! empty( $data['alt'] ) ? $data['alt'] : '';
+
+		return new static( $source, $alt );
 	}
 }
--- a/includes/image/BaseImage.php
+++ b/includes/image/BaseImage.php
@@ -1,0 +1,33 @@
+<?php
+
+class BaseImage implements WPNotify_Image, WPNotify_JsonUnserializable {
+
+	/**
+	 * It can be and URL of an image file or and data/image source
+	 * @var string source for "src" attribute of an <img/> html node
+	 */
+	protected $source;
+
+	/**
+	 * @var string for alternative text of an <img/> html node
+	 */
+	protected $alt = '';
+
+	/**
+	 * BaseImage constructor.
+	 *
+	 * @param string $source
+	 * @param string $alt
+	 */
+	public function __construct( $source, $alt = '' ) {
+		$this->source = $source;
+	}
+
+	public function jsonSerialize() {
+		// TODO: Implement jsonSerialize() method.
+	}
+
+	public static function json_unserialize( $json ) {
+		// TODO: Implement json_unserialize() method.
+	}
+}
--- a/includes/image/BaseImage.php
+++ b/includes/image/BaseImage.php
@@ -1,6 +1,6 @@
 <?php
 
-class BaseImage implements WPNotify_Image, WPNotify_JsonUnserializable {
+class WPNotify_BaseImage implements WPNotify_Image, WPNotify_JsonUnserializable {
 
 	/**
 	 * It can be and URL of an image file or and data/image source
@@ -21,10 +21,18 @@ class BaseImage implements WPNotify_Image, WPNotify_JsonUnserializable {
 	 */
 	public function __construct( $source, $alt = '' ) {
 		$this->source = $source;
+		$this->alt    = $alt;
 	}
 
+	/**
+	 * @return array
+	 */
 	public function jsonSerialize() {
-		// TODO: Implement jsonSerialize() method.
+
+		return array(
+			'source' => $this->source,
+			'alt'    => $this->alt
+		);
 	}
 
 	public static function json_unserialize( $json ) {
--- a/includes/image/BaseImage.php
+++ b/includes/image/BaseImage.php
@@ -1,6 +1,6 @@
 <?php
 
-class WPNotify_BaseImage implements WPNotify_Image, WPNotify_JsonUnserializable {
+class WPNotify_BaseImage implements WPNotify_Image {
 
 	/**
 	 * It can be and URL of an image file or and data/image source
@@ -29,10 +29,17 @@ class WPNotify_BaseImage implements WPNotify_Image, WPNotify_JsonUnserializable 
 	 */
 	public function jsonSerialize() {
 
-		return array(
-			'source' => $this->source,
-			'alt'    => $this->alt
-		);
+		$data = array();
+
+		if ( ! empty( $this->get_source() ) ) {
+			$data['source'] = $this->get_source();
+		}
+
+		if ( ! empty( $this->get_alt() ) ) {
+			$data['alt'] = $this->get_alt();
+		}
+
+		return $data;
 	}
 
 	/**
@@ -65,6 +72,6 @@ class WPNotify_BaseImage implements WPNotify_Image, WPNotify_JsonUnserializable 
 		$source = ! empty( $data['source'] ) ? $data['source'] : '';
 		$alt    = ! empty( $data['alt'] ) ? $data['alt'] : '';
 
-		return new static( $source, $alt );
+		return new self( $source, $alt );
 	}
 }
--- a/includes/image/Image.php
+++ b/includes/image/Image.php
@@ -1,0 +1,5 @@
+<?php
+
+interface WPNotify_Image extends JsonSerializable {
+
+}
--- a/includes/image/Image.php
+++ b/includes/image/Image.php
@@ -1,5 +1,5 @@
 <?php
 
-interface WPNotify_Image extends JsonSerializable {
+interface WPNotify_Image extends JsonSerializable, WPNotify_JsonUnserializable {
 
 }
--- a/includes/senders/BaseSender.php
+++ b/includes/senders/BaseSender.php
@@ -27,10 +27,16 @@ class WPNotify_BaseSender implements WPNotify_Sender, WPNotify_JsonUnserializabl
 	 * @return array
 	 */
 	public function jsonSerialize() {
-		return array(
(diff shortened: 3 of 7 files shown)


includes/image/BaseImage.php
```diff
@@ -0,0 +1,70 @@
...
+	 * @return WPNotify_JsonUnserializable
+	 */
+	public static function json_unserialize( $json ) {
+
+		$data = json_decode( $json, true );
+
+		$source = ! empty( $data['source'] ) ? $data['source'] : '';
+		$alt    = ! empty( $data['alt'] ) ? $data['alt'] : '';
+
+		return new static( $source, $alt );
```
raaaahman: I notice that sometimes we use `new self( ...$args );` and sometimes we use ` new static( ...$args );`. Which one better suits our needs in your opinion?

tests/phpunit/tests/base-image.php
```diff
@@ -0,0 +1,79 @@
+<?php
+
+class Test_WPNotify_BaseImage extends WPNotify_TestCase {
+
+	/**
+	 * @param array  $image_data
+	 * @param string $expected_json
+	 *
+	 * @dataProvider data_provider_images
```
raaaahman, line 9: Good idea to use the same data provider for both tests! :+1:

tests/phpunit/tests/base-image.php
```diff
@@ -0,0 +1,79 @@
...
+
+		list( $source, $alt ) = array_values( $image_data );
+
+		$instance = new WPNotify_BaseImage( $source, $alt );
+
+		$this->assertEquals( $instance->get_source(), $test_instance->get_source() );
+		$this->assertEquals( $instance->get_alt(), $test_instance->get_alt() );
+	}
+
+	public function data_provider_images() {
```
raaaahman, line 42: Your four cases provide a good test coverage for our code. Well done!

As a side note, you can name each individual case and PHPUnit will use these names in its failure reports. For example:

```php
       'With image source and alternative text' => array(
		array(
			'source' => 'img-source1',
			'alt'    => 'img-alt1',
		),
		'{"source":"img-source1","alt":"img-alt1"}',
	),

```

includes/image/Image.php
```diff
@@ -0,0 +1,5 @@
+<?php
+
+interface WPNotify_Image extends JsonSerializable {
```
raaaahman: This class also validates the `WPNotify_JsonUnserializable` interface.

includes/senders/BaseSender.php
```diff
@@ -36,16 +44,26 @@ public static function json_unserialize( $json ) {
 
 		$data = json_decode( $json, true );
 
-		$name = ! empty( $data['name'] ) ? $data['name'] : '';
+		$name  = ! empty( $data['name'] ) ? $data['name'] : '';
+		$image = ! empty( $data['image'] ) ? WPNotify_BaseImage::json_unserialize( json_encode( $data['image'] ) ) : null;
```
raaaahman: What happens if we set the `$associative` parameter of `json_decode()` to false? Do we have our properties correctly instantiated as object? 

If this doesn't work, could we use `$depth` parameter of the `json_decode()` function to only decode the first level objects? We may able able to have our image as a JSON string and then not have to encode it again.

https://www.php.net/manual/fr/function.json-decode.php

tests/phpunit/tests/base-image.php
```diff
@@ -0,0 +1,79 @@
...
+				),
+				'{"source":"img-source1","alt":"img-alt1"}',
+			),
+
+			array(
+				array(
+					'source' => 'img-source2',
+					'alt'    => '',
+				),
+				'{"source":"img-source2","alt":""}',
```
raaaahman: Do you think we should encode empty properties in our JSON strings? Maybe we could only encode the non empty instead.

For example, the expected JSON here would be  `'{"source":"img-source2"}`.

tests/phpunit/tests/base-sender.php
```diff
@@ -3,38 +3,75 @@
...
-		$senderInstance = new WPNotify_BaseSender( $sender_name );
-		$encoded        = json_encode( $senderInstance );
-		$this->assertEquals( $expected_json, $encoded );
+		$sender_instance = null;
+
+		if ( is_string( $sender ) ) {
+			$sender_instance = new WPNotify_BaseSender( $sender );
+		} else if ( is_array( $sender ) ) {
+			$sender_instance = new WPNotify_BaseSender( $sender['name'], $sender['image'] );
+		}
```
raaaahman: As a rule of thumb, I'd rather not have conditional logic in our tests: the less we have to think about what the test verifies, the quicker we could read them.

Maybe you could instantiate these objects in the `@dataProvider` instead. Or maybe you could use two different tests with no provider.

tests/phpunit/tests/base-notification.php
```diff
@@ -2,7 +2,7 @@
 
 class Test_WPNotify_BaseNotification extends WPNotify_TestCase {
 
-	const JSON_SERIALIZED = '{"WPNotify_RecipientCollection":[],"WPNotify_BaseMessage":"Message","WPNotify_BaseSender":{"name":"Name 1"}}';
+	const JSON_SERIALIZED = '{"WPNotify_RecipientCollection":[],"WPNotify_BaseMessage":"Message","WPNotify_BaseSender":{"name":"Name 1","image":null}}';
```
raaaahman: Same comment than for the `Test_WPNotify_BaseImage::data_provider_images()`: maybe we don't want the empty `"image"` field to be printed in our JSON string. :thinking:

includes/image/BaseImage.php
danbilauca: I'd say `static` even though I could see all over the code `self`. I might have missed this one.

tests/phpunit/tests/base-notification.php
danbilauca: fixed

tests/phpunit/tests/base-sender.php
danbilauca: fixed

tests/phpunit/tests/base-image.php
danbilauca: fixed

includes/senders/BaseSender.php
danbilauca: I've changed the way the image is initialised. I've used the same logic from `WPNotify_BaseNotification` instantiating the image from reflection.

includes/image/Image.php
danbilauca: fixed

tests/phpunit/tests/base-image.php
danbilauca: fixed

includes/image/BaseImage.php
danbilauca: fixed

includes/image/BaseImage.php
```diff
@@ -1,6 +1,6 @@
 <?php
 
-class WPNotify_BaseImage implements WPNotify_Image, WPNotify_JsonUnserializable {
```
raaaahman: I didn't see the `WPNotify_JsonUnserializable` interface use here, sorry. Anyway, it is better how it is declared right now.

includes/senders/BaseSender.php
```diff
@@ -27,10 +27,16 @@ public function __construct( $name, $image = null ) {
...
+
+		$data = array(
+			'name' => $this->name,
 		);
+
+		if ( $this->image ) {
+			$data[ get_class( $this->image ) ] = $this->image;
+		}
+
+		return array_filter( $data );
```
raaaahman: I am not sure what the `array_filter` function is supposed to do without passing it a callback function. Would you explain its usage here?

tests/phpunit/tests/base-sender.php
```diff
@@ -3,74 +3,67 @@
...
 		/**
 		 * If an Image has been sent
 		 */
 		if ( $testee->get_image() ) {
-			$this->assertInstanceOf( 'WPNotify_BaseImage', $testee->get_image() );
-			$this->assertEquals( $sender_instance->get_image()->get_source(), $testee->get_image()->get_source() );
-			$this->assertEquals( $sender_instance->get_image()->get_alt(), $testee->get_image()->get_alt() );
+			$this->assertInstanceOf( get_class( $sender->get_image() ), $testee->get_image() );
+			$this->assertEquals( $sender->get_image()->get_source(), $testee->get_image()->get_source() );
+			$this->assertEquals( $sender->get_image()->get_alt(), $testee->get_image()->get_alt() );
```
raaaahman: If line 38 passes, is there any way for lines 45 to 46 to fail?

I mean if `$sender->get_image() === $testee->get_image()`, then their classes, `source` property and `alt` property probably match as well.

But  `$sender->get_image() === $testee->get_image()` imply that the two `image` properties refer to the same instance... Is this something we intended here?

tests/phpunit/tests/base-sender.php
```diff
@@ -3,74 +3,67 @@
...
-			array(
+			'sender with image' => array(
 				array(
-					'name'  => 'Name 2',
-					'image' => new WPNotify_BaseImage( 'img-source', 'img-alt' ),
+					'name'               => 'Name 2',
+					'WPNotify_BaseImage' => new WPNotify_BaseImage( 'img-source', 'img-alt' ),
 				),
-				'{"name":"Name 2","image":{"source":"img-source","alt":"img-alt"}}',
+				'{"name":"Name 2","WPNotify_BaseImage":{"source":"img-source","alt":"img-alt"}}',
```
raaaahman: The Json serialization was meant to be used to communicate between the back-end and front-end parts of the feature.

I am not sure what the front-end part would do with a PHP class name. Wouldn't this be bothersome to have to change the data field names on the front-end if we ever want to add a new PHP class for images?

raaaahman (commented):
Nice to have added explicit names to the test cases in the data providers.

I do not really understand the use of `ReflectionClass` here, is that because you wanted to pass parameter to the constructor as an array?

tests/phpunit/tests/base-sender.php
raaaahman: PHPUnit does not perform a deep comparison between the two `WPNotify_BaseImage` objects, so we need to compare their individual properties.

tests/phpunit/tests/base-sender.php
raaaahman: This is already the way `WPNotify_BaseNotifications` are instantiated, it's better to keep our code consistent.

raaaahman (approved):

PR #53, opened 2021-04-14
Adapt existing code to WordPress' coding standards

raaaahman:
## Goal

Make previously written code compliant to WordPress Coding Standards.

## Changes

- Names of class use Snake_Case formatting
- Names of files use kebab-case formatting, appended with either `class` or `interface` (the latter is not a WordPress Coding Standards, although it seemed fitting)
- Various spacings and indentation automatic fixes


jonathanbossenger (approved):
LGTM 👍

PR #54, opened 2021-04-21
Add PHP linting and unit tests to a GitHub workflow

raaaahman:
## Goal

This PR aims to automatically check the WordPress Code Standards and run our PHP unit tests each time a pull request is made on our `develop` and `trunk` branches.

See #10 

- Inspired by [woccommerce/woocommerce](https://github.com/woocommerce/woocommerce/blob/trunk/.github/workflows/pr-unit-tests.yml)
- Uses [shivammathur/setup-php](https://github.com/shivammathur/setup-php)
- Runs on Ubuntu 18.04 to benefit from pre-installed PHP versions, see: https://github.com/shivammathur/setup-php#tada-php-support
- #53 Needs to be merged first for our code sniffes to pass

## Running locally

The GitHub actions can be run locally using [nektos/act](https://github.com/nektos/act) and [shivammathur/node-docker](https://github.com/shivammathur/node-docker). See explanation: https://github.com/shivammathur/setup-php#local-testing-setup

Run the command

```bash
act -P ubuntu-18.04=shivammathur/node:bionic pull_request
```

## WIP

- [x] ~~Can't run the action locally using [nektos/Act](https://github.com/nektos/act): The composer package appears missing (medium install)~~ 
- [x] ~~The action runs on the server but breaks because our set of dependencies are not compatible with PHP 8~~
- [x] Could be useful to set a matrix of PHP versions to run the unit tests with. **For the moment only runs on PHP 7.2, 7.3 and 7.4.**
- [x] The linting can be separated from the PHP versions matrix.
- [ ] Won't be enough to run integration tests, because it needs a setup connecting a MySQL database and a WordPress install. See how [WordPress/wordpress-develop](https://github.com/WordPress/wordpress-develop/blob/master/.github/workflows/phpunit-tests.yml) and [WordPress/gutenberg](https://github.com/WordPress/gutenberg/blob/bee52e68292357011a799f067ad47aa1c1d710e1/.github/workflows/unit-test.yml#L52) manage it.
- [x] Coding standards should show as failed... when failing

Issue #10, opened 2020-09-07
Set up Continuous Integration


raaaahman:
Since we are hosting the plugin code on GitHub, and this is an open source project, [Travis CI](https://github.com/marketplace/travis-ci/plan/MDIyOk1hcmtldHBsYWNlTGlzdGluZ1BsYW43MA==#pricing-and-setup) seems to be the way to go.

The benefits of continuous integration will be to check on each PR:

- The passing or failing of our automated tests suite (see #12, #13 and #15 )
- The respect of coding standards (see #11 )
- The compatibility with different server environments (see #14 )

szepeviktor:
What about **superior** [GitHub Actions](https://github.com/marketplace?type=actions&query=WordPress)?

szepeviktor:
this is what I call "3 basic tests" - turns out they are for **elite** developers

1. Lint source code (syntax check) - see `php -h | grep Syntax` why it is called "lint"
1. Check coding style and adhere to EditorConfig
1. Do static analysis

[The newcomer](https://github.com/szepeviktor/debian-server-tools/blob/master/webserver/Continuous-integration-Continuous-delivery.md) - Number Zero: :zero: _Check non-ASCII characters in the source code_

jonathanbossenger:
I would love to see this project make use of GitHub actions, as a model for how an open source WordPress project can leverage their benefits.

youknowriad:
Gutenberg is already 100% Github actions powered. you might want to use some of the jobs there.

jonathanbossenger:
@raaaahman it might be worthwhile to review what Gutenberg already does, and see how much of that we can use.

raaaahman:
The [WordPress Core team has worked on GitHub actions](https://make.wordpress.org/core/2020/10/15/introducing-github-actions-for-automated-testing/) as well, we'd probably can draw inspiration from there too.

erikyo:
Has been already setup :) 

duplicated issue of #54, #59...

PR #59, opened 2021-06-15
Run Unit Tests and Coding Standards on Pull Request

raaaahman:
## Goal

This PR sets up a workflow for Continuous Integration of our PHP code using GitHub Actions.

Each time a pull request is made on our `develop` and `trunk` branches:

-  automatically check the WordPress Code Standards 
-  run our PHP unit tests 

See #10 
Replaces #54 

## Details

- Inspired by [woccommerce/woocommerce](https://github.com/woocommerce/woocommerce/blob/trunk/.github/workflows/pr-unit-tests.yml)
- Uses [shivammathur/setup-php](https://github.com/shivammathur/setup-php)
- Uses [WPCS GitHub Action](https://github.com/10up/wpcs-action) from 10up (thanks @dinhtungdu)
- Uses the `install-wp-tests.sh` script created by the [wp scaffold plugin-tests]( [wp scaffold plugin-tests](https://developer.wordpress.org/cli/commands/scaffold/plugin-tests/) ) command.

## Note

You may be able to run the Github Actions on your computer with [act](https://github.com/nektos/act), but you probably need the full installation of `nektos/act` to have SVN available... (takes lots of disk space)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -18,6 +18,7 @@
     "minimum-stability": "dev",
     "prefer-stable": true,
     "scripts": {
-        "lint": "vendor/squizlabs/php_codesniffer/bin/phpcs --report=summary,source"
+        "lint": "vendor/squizlabs/php_codesniffer/bin/phpcs --report=summary,source",
+	"test": "vendor/bin/phpunit"
     }
 }
--- a/composer.json
+++ b/composer.json
@@ -18,7 +18,7 @@
     "minimum-stability": "dev",
     "prefer-stable": true,
     "scripts": {
-        "lint": "vendor/squizlabs/php_codesniffer/bin/phpcs --report=summary,source",
-	"test": "vendor/bin/phpunit"
+        "lint": "vendor/bin/phpcs",
+        "test": "vendor/bin/phpunit"
     }
 }
--- a/tests/bin/install-wp-tests.sh
+++ b/tests/bin/install-wp-tests.sh
@@ -1,0 +1,155 @@
+#!/usr/bin/env bash
+
+if [ $# -lt 3 ]; then
+	echo "usage: $0 <db-name> <db-user> <db-pass> [db-host] [wp-version] [skip-database-creation]"
+	exit 1
+fi
+
+DB_NAME=$1
+DB_USER=$2
+DB_PASS=$3
+DB_HOST=${4-localhost}
+WP_VERSION=${5-latest}
+SKIP_DB_CREATE=${6-false}
+
+TMPDIR=${TMPDIR-/tmp}
+TMPDIR=$(echo $TMPDIR | sed -e "s/\/$//")
+WP_TESTS_DIR=${WP_TESTS_DIR-$TMPDIR/wordpress-tests-lib}
+WP_CORE_DIR=${WP_CORE_DIR-$TMPDIR/wordpress/}
+
+download() {
+    if [ `which curl` ]; then
+        curl -s "$1" > "$2";
+    elif [ `which wget` ]; then
+        wget -nv -O "$2" "$1"
+    fi
+}
+
+if [[ $WP_VERSION =~ ^[0-9]+\.[0-9]+\-(beta|RC)[0-9]+$ ]]; then
+	WP_BRANCH=${WP_VERSION%\-*}
+	WP_TESTS_TAG="branches/$WP_BRANCH"
+
+elif [[ $WP_VERSION =~ ^[0-9]+\.[0-9]+$ ]]; then
+	WP_TESTS_TAG="branches/$WP_VERSION"
+elif [[ $WP_VERSION =~ [0-9]+\.[0-9]+\.[0-9]+ ]]; then
+	if [[ $WP_VERSION =~ [0-9]+\.[0-9]+\.[0] ]]; then
+		# version x.x.0 means the first release of the major version, so strip off the .0 and download version x.x
+		WP_TESTS_TAG="tags/${WP_VERSION%??}"
+	else
+		WP_TESTS_TAG="tags/$WP_VERSION"
+	fi
+elif [[ $WP_VERSION == 'nightly' || $WP_VERSION == 'trunk' ]]; then
+	WP_TESTS_TAG="trunk"
+else
+	# http serves a single offer, whereas https serves multiple. we only want one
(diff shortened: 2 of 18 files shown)


tests/bin/install-wp-tests.sh
```diff
@@ -0,0 +1,155 @@
...
+			EXTRA=" --host=$DB_HOSTNAME --port=$DB_SOCK_OR_PORT --protocol=tcp"
+		elif ! [ -z $DB_SOCK_OR_PORT ] ; then
+			EXTRA=" --socket=$DB_SOCK_OR_PORT"
+		elif ! [ -z $DB_HOSTNAME ] ; then
+			EXTRA=" --host=$DB_HOSTNAME --protocol=tcp"
+		fi
+	fi
+
+	# create database
+	mysqladmin create $DB_NAME --user="$DB_USER" --password="$DB_PASS"$EXTRA
```
JasonTheAdams, line 150: I recommend replacing this with the following. The reason being that the default WordPress `install.sh` _does not_ play nicely with spaces in paths. So if I did `install-wp-tests.sh root wp wp "/some/path/Local Sites/site"` it would not work due to the space in "Local Sites". This can also occur in the socket path. The following is modified to handled said spaces.
```suggestion
	# parse DB_HOST for port or socket references
	IFS=':'; read -r -a PARTS <<< "$DB_HOST"; unset IFS;

	local DB_HOSTNAME=${PARTS[0]};
	local DB_SOCK_OR_PORT=${PARTS[1]};
	local EXTRA=""

	if [ -n "$DB_HOSTNAME" ] ; then
		if echo "$DB_SOCK_OR_PORT" | grep -q '^[0-9]\{1,\}$' ; then
			EXTRA="--host=$DB_HOSTNAME --port=$DB_SOCK_OR_PORT --protocol=tcp"
		elif [ -n "$DB_SOCK_OR_PORT" ] ; then
			EXTRA="--socket=\"${DB_SOCK_OR_PORT}\""
		elif [ -n "$DB_HOSTNAME" ] ; then
			EXTRA="--host=$DB_HOSTNAME --protocol=tcp"
		fi
	fi

	# create database
	eval "mysqladmin create ${DB_NAME} --user=${DB_USER} --password=${DB_PASS} ${EXTRA}"
```

raaaahman: Thank you for your suggestion.

My bad, I should have stated in the PR description that I didn't wrote the `install-wp-tests.sh` script, but rather copied the one outputted by the [wp scaffold plugin-tests](https://developer.wordpress.org/cli/commands/scaffold/plugin-tests/) command.

Do you believe your suggestion would make sense to be reported to the WP CLI team? I'd rather keep this script consistent with WP CLI, but if you feel that our case would need a specific behavior that is not applicable to every plugin, I'd be interested to hear it.

dinhtungdu:
Hi @raaaahman, about the PHPCS check, I reccommend using a PHPCS action for annotation support. GitHub annotation provides a better way to display linting errors and warning right from the Files tab of PRs. It also saves us some action minutes because we don't need to install composer packages or set up PHP.

You can check out the action here: https://github.com/chekalsky/phpcs-action.

I forked the action above and add support for WordPress Coding Standards here: https://github.com/10up/wpcs-action. It hasn't supported custom PHPCS rulesets yet but I have a plan to add that feature. If we only use WPCS, it's ready to use now.



<details><summary>GitHub annotation example.</summary>

image

</details>

composer.json
```diff
@@ -18,6 +18,7 @@
     "minimum-stability": "dev",
     "prefer-stable": true,
     "scripts": {
-        "lint": "vendor/squizlabs/php_codesniffer/bin/phpcs --report=summary,source"
+        "lint": "vendor/squizlabs/php_codesniffer/bin/phpcs --report=summary,source",
+	"test": "vendor/bin/phpunit"
```
dinhtungdu: Indentation is mixed here. Tab is used for the 22nd line.
phpcs path can be shortened to `vendor/bin/phpcs` as well.

raaaahman: > Hi @raaaahman, about the PHPCS check, I reccommend using a PHPCS action for annotation support.

That could be pretty neat indeed!

tests/bin/install-wp-tests.sh
jonathanbossenger: @JasonTheAdams as @raaaahman has pointed out, this seems to be a bug in the [WP-CLI](https://github.com/wp-cli/wp-cli) `install-wp-tests.sh` script. Would you be able to report this as an issue there, or even submit a PR?

JasonTheAdams: Sure! PR submitted: https://github.com/wp-cli/scaffold-command/pull/291

jonathanbossenger: Thanks, @JasonTheAdams

For now, until the PR is merged into WP-CLI, are you happy with us proceeding to merge this PR as is? I know it doesn't solve the immediate problem, but once WP-CLI is updated, we'll get the changes downstream.

JasonTheAdams: Yeah, if you want to stay in parity with WP CLI — which I understand — then I would just proceed. No need to hold things up for this. The code can be updated at another point. Hopefully this doesn't come up too often for folks.

erikyo:
Is it possible to continue/merge this? It would also be interesting because we could also add an automated build for node which avoids publishing the "build/" folder.

erikyo:
@raaaahman, i've [rebased your very useful PR into the development branch](https://github.com/erikyo/wp-notify/tree/tool/github-actions-merge) but look like the [coding standards check fails](https://github.com/erikyo/wp-notify/actions/runs/2476466367) due the file autogenerated by webpack that needs to be excluded. Please could you take a look? thanks!

raaaahman:
Sorry @erikyo , your comment went totally under my radar. To be honest, I don't use much WordPress nor PHP these days.

If fortunate, all it would take to solve your CS issue would be to add an exclude pattern in the configuration file: https://github.com/WordPress/wp-feature-notifications/blob/117445ba3357df5c0a456c03841e1f0d67e0d597/phpcs.xml.dist#L50

I can't see your linked job logs, so I don't know the exact pattern this needs to be. Let me know if you have further issues with this.

erikyo:
yes it was exactly like that... among the exclusions the `/build` folder was missing... now I think we have fixed it, thank you anyway!

Issue #60, opened 2021-06-29
Update Notification model with optional data


raaaahman:
The actual implementation of `WP_Notify_Base_Notification` is not up-to-date with the database scheme merged in #21 .

Although, all the missing fields are not required, but optional ones. We should find a mechanism to extend the `WP_Base_Notification` model without having to modify it from the inside (which would then need to propagate those changes to all classes that uses this model).

PR #61, opened 2021-07-06
Extend notification model with optional properties.

raaaahman:
## Before

The `WP_Notify_Base_Notification` model was not up-to-date with the optional fields defined in PR #21

See #60 

## Changes

- Introduces the `title`, `image` and `action_link`  properties as decorators to the `WP_Notify_Base_Notification` class.
- Uses JSON as test data, which should be reused when testing for displaying these notifications on the front-end.
- Adds the `image` field in the database scheme.

## Current limitations

The unserialization of the decorated notifications is not ready yet. It is debatable if this would be used at all (receiving notification data from the front-end), but on another hand, it would be better to keep this possibility open at such an early stage of development...


raaaahman:
Replaced by #63

PR #66, opened 2022-03-08
Feature/notification reworked design

erikyo:
- The design has been updated
- Adds webpack 

**Folder structure:**
includes/ui/notification-hub/assets - The work folder
assets/html - here you will find the notices and the sidebar html
assets/script - here for now you will find some scripts that has only the function to test the ui
assets/css/notify.scss - the WP-Notify style (this plugin css)
assets/css/wordpress.scss - The WordPress Admin style (will not be included in the future, but at the moment is needed for ui testing purpose)
img/ - actually there is the favicon (without that there is an error in console)
includes/ui/notification-hub/dist - Build folder


jonathanbossenger:
Thanks @erikyo I will work on reviewing it over the course of this week.

.browserslistrc
```diff
@@ -0,0 +1,2 @@
+defaults
```
Sephsekla: If we are doing a proper build process, I'd recommend we use the WP module for this that way we know we're getting the same as core. https://developer.wordpress.org/block-editor/reference-guides/packages/packages-browserslist-config/

erikyo: Good point! it's important that the settings are the same as for the core so that there will be no discrepancies when we take it out of webpack... will fix

.gitignore
```diff
@@ -1 +1,11 @@
-.DS_Store
```
Sephsekla, line 1: I'd recommend keeping this to avoid committing random mac files 😁

jonathanbossenger:
@Sephsekla from my POV this looks good. If you have time to look this over, please do.

Alternatively, if you don't have time for a full review, it would be helpful to compare what you've done to this PR, and confirm that if we merge this, it compliments your work. From what I can see the folder structure is slightly different, in that you've built your work on the notification-hub directrory, and this PR builds everything in the docs directory.

I just want to make sure we've not lost anything important you included in your branch.

jonathanbossenger:
@erikyo thanks for your patience here, it's been a busy few months.

So, I have a few things I'd like to see if we can get implemented into this PR, in order to make it more easily "testable" by end users.

- Currently for the user to view this, they will need to know to navigate to /wp-content/plugins/wp-notify/docs/index.html. I understand why this is, but I'd like to make it easier for someone to install the plugin and find the designs. Would you be able to register a custom Admin menu (maybe something like WP Notify) that would direct the user to this URL instead?
- At the moment the "Try this new feature" buttons slide open the Notifications hub. Could we rather change that so that it opens the link to this repository https://github.com/WordPress/wp-notify in a new tab?
- The HTML title reads `<title>WordPress Dashboard + &lt;%= htmlWebpackPlugin.options.title %&gt;</title>` could we fix that so that instead of `&lt;%= htmlWebpackPlugin.options.title %&gt;` it says `WP Notify`?

If we could make those changes, it makes it easier to share this as an installable plugin, which users can install and test. 

Let me know if you need any help with any of these items.

erikyo:
All clear but I just have one small question:

> Would you be able to register a custom Admin menu (maybe something like WP Notify) that would direct the user to this URL instead? 

Where i have to register the custom Admin menu in the wp-notify plugin ([here?](https://github.com/WordPress/wp-notify/blob/develop/includes/class-wp-notify-factory.php)). Let me know!

jonathanbossenger:
@erikyo for now, I suggest we just add the admin menu procedurally to the main wp-notify.php file. This admin menu will be removed when we actually develop the notifications functionality, it's just to allow folks to install and see this in action, and hopefully generate some interest.

README.md
```diff
@@ -3,3 +3,24 @@
 A feature plugin for WordPress, which aims to create a new (better) way to manage and deliver notifications to the relevant audience.
 
 To get started, please take a look at our [project wiki](https://github.com/WordPress/wp-notify/wiki)
+
+## Useful commands
+
+In order to run the following commands, you need to have [Node.js](https://nodejs.org) (including `npm`)
```
Sephsekla, line 9: For ease of reproducing any bugs, what do you think about having a specific node/npm version here? Ideally I'd recommend using `nvm` and a `.nvmrc` file.

includes/ui/notification-hub/assets/script/script.js
```diff
@@ -0,0 +1,55 @@
...
+  document.querySelectorAll('.wp-notification-hub-trigger').forEach((e) => {
+    e.addEventListener("click", function() {
+      hub.classList.toggle("active")
+    });
+  })
+
+  // Click handler to add a new notification
+  document.getElementById('save-post').addEventListener("click", function(e) {
+    e.preventDefault();
+    let title = document.getElementById('title').value;
```
Sephsekla, line 19: I think these two `let`s can be `const`s, right?

includes/ui/notification-hub/assets/script/script.js
```diff
@@ -0,0 +1,55 @@
...
+    let content = document.getElementById('content').value;
+
+    var div  = document.createElement('div');
```
Sephsekla, line 22: Ditto this can be a `const`, since we're just mutating properties, rather than ever setting it again.

Sephsekla, line 22: Ditto this can be a `const`, since we're just mutating properties, rather than ever setting it again.

includes/ui/notification-hub/assets/style/notify/bullet.scss
```diff
@@ -0,0 +1,30 @@
...
+    top: calc(50% - #{$bullet-width * .5 });
+
+    border-radius: 50%;
+    background: var(--wp-notify_color-error);
+  }
+}
+
+#wpadminbar .wp-notification.unread >:first-child:before {
+  top: calc(50% - #{$bullet-width * .5 });
+  left: #{ -24px * .5 - $bullet-width * .5 }; // 9 - left padding - border /2  + bullet/2
```
Sephsekla, line 25: These seem really fiddly, I'm wondering if there's a clearer way to handle the variables etc here.

includes/ui/notification-hub/assets/style/notify/notification-hub.scss
```diff
@@ -0,0 +1,212 @@
...
+  position: fixed;
+  right: calc(0px - var(--wp-notify_hub-width));
+  top: 0;
+  z-index: 100001;
+  display: flex;
+  flex-direction: column;
+  width: 100%;
+  height: 100%;
+  max-width: var(--wp-notify_hub-width);
+  padding: 30px 0 50px;
```
Sephsekla, line 14: Wondering if longer-term we should investigate use of `em`/`rem` rather than exact pixel values.

includes/ui/notification-hub/assets/style/notify/notify.scss
```diff
@@ -0,0 +1,83 @@
+/*
+ * General notifications style
+ */
+.wp-notification {
```
Sephsekla, line 4: I'm conscious we have styles for the individual notifications in both this file and the hub file - not necessarily a bad thing but I do wonder if it could be slightly unclear for future maintenance.

webpack.config.js
```diff
@@ -0,0 +1,188 @@
+const isProduction = process.env.NODE_ENV === 'production';
```
Sephsekla, line 1: I'd maybe split this file out and import a couple of things, but that's a matter of personal preference more than anything.

includes/ui/notification-hub/assets/html/notices.html
```diff
@@ -0,0 +1,50 @@
+<div class="wp-notification wp-notice-alert is-dismissible">
```
Sephsekla, line 1: Given you've already got the html `<include>` setup for use elsewhere, what do you think about splitting these out into smaller files in a subfolder and including them? Just for the purpose of keeping things as modular as possible and easy to see what's what.

Sephsekla (commented):
Thanks for all the hard work here, this has really got things moving again!

I haven't yet tested it out, so just a general scan of the code, but I've added a few comments and suggestions. Some are really nit-picky, but I've just added things that have come to mind as I've read through.

includes/ui/notification-hub/assets/style/notify/notify.scss
erikyo: I removed from here all the css properties apart from colors and lettering (which are indeed shared between the two types of notifications), tell me if this is an acceptable solution for you.

webpack.config.js
erikyo: here the issue was solved because webpack is no longer used as long we don't need no more the dev server 😉... checkout at [this branch](https://github.com/erikyo/wp-notify/tree/design_implementation) where I am integrating this style (with all your suggestions)

includes/ui/notification-hub/assets/style/notify/bullet.scss
erikyo: I honestly don't know how to solve it here, but it can be simplified with #{ (-24px - $bullet-width) * .5 }

jonathanbossenger:
Superceeded by #67

Issue #69, opened 2022-06-01
Update the Plugin Name value


jonathanbossenger:
Can we update the Plugin Name in the main plugin file to `WP Feature Notifications`. This will resolve the issue of it conflicting with the other plugin in the WP repository?

PR #70, opened 2022-06-01
#69 update plugin name

Sephsekla:
#69 Updates the plugin name and also the `composer.json` name field.


erikyo:
another way to fix could be to add the version > 1.0.1 in this way
```
/**
 * Plugin Name:       WP Notify
 * Plugin URI:        URI
 * Description:       DESC
 * Version:           1.0.1
 * Requires at least: 6.0
 * Requires PHP:      7.1.0
 * Author:            AUTHOR
 * Author URI:        AUTHOR_URI
 * License:           GPL v2 or later
 * License URI:       https://www.gnu.org/licenses/gpl-2.0.html
 * Text Domain:       wp-notify
 */
```

Sephsekla:
I'd personally rather avoid any ambiguity at all, since they're not remotely the same plugin.

@jonathanbossenger what do you think?

jonathanbossenger:
Just thinking out loud, should we also update the composer.json name to `wordpress/wp-feature-notifications`.

WP Notify was just a generica name we chose for the project, but WP Feature Notifications more accurately describes the project and prevents any future conflicts. It also ties in nicely with our Slack channel name.

Sephsekla:
That works for me, I'll update this a bit later on today.

jonathanbossenger:
While we're at it, we might as well rename the main plugin file to `wp-feature-notifications.php` as well

Sephsekla:
Sure, l'll handle this and see if there's anything else that will want changing in line with this.

Sephsekla:
Sorry all, I've been busy with flights etc this week. I'll try and get this sorted shortly.

jonathanbossenger (approved):

PR #73, opened 2022-06-02
Avoid issues with phpcs

erikyo:
The fully empty file creates troubles with phpcs (and conseguently with commit hooks)... I created an empty class into "includes/restapi/class-wp-notify-notification-controller.php" to overcome the issue


bacoords (approved):
Just tested locally and all good.

erikyo:
@jonathanbossenger can this be merged? If not, it is not possible to contribute (unless you set the flag "--no-verify")

PR #74, opened 2022-06-02
Disables notification hub visibility clicking on page body

erikyo:
FIX: https://github.com/WordPress/wp-feature-notifications/pull/67#issuecomment-1143089828

> Does anyone else think that clicking outside of the drawer-style sidebar should make it disappear? That's a pretty common pattern and I don't see any other UI element to close the sidebar. (@bacoords)

> Personally as a user, that's definitely the behaviour I'd expect it to have. (@Sephsekla)

Creates a click event listener on body when the drawer is open, when triggered it closes the drawer and remove the listener


bacoords (approved):
This works for me - I had to build it locally.

PR #75, opened 2022-06-02
Updates the browserlist configuration package.json

erikyo:
Drops the IE support, that isn't needed


bacoords (commented):
This worked to remove the warning. 

How do we feel about using this: https://github.com/WordPress/gutenberg/tree/trunk/packages/browserslist-config

bacoords (approved):
Nice! This worked for me

PR #77, opened 2022-06-08
Adds the "version number" constant

erikyo:
Will set the right version to scripts and styles, at the moment there is an hardcoded string "VERSION" that could lead to warning in console


bacoords (approved):
I love a good version number constant. This looks good to me.

jonathanbossenger:
> I love a good version number constant. This looks good to me.

Best PR review comment ever!

PR #82, opened 2022-06-12
Tool/GitHub actions

erikyo:
followup #59

updates #59 in order to solve conflicts


Sephsekla:
I'm not necessarily saying do this, but I'd be interested to hear your thoughts on using `phpunit` with the `testdox` flag, that way all tests would be listed for completeness if you needed to go through the CI log.

Useful, or just extra noise?

Sephsekla:
It would also be good (probably in a followup) to expand this to eslint and stylelint given we're not just doing PHP work.

erikyo:
@Sephsekla you are right and testdox may be useful but the goal of this PR anyway is to get the CI working with the basic checks. In the future we surely need expand that topic, but right now I don't see all this need since the tests are yet missing.

I think the lint needs to be done before the commit (and we may want to add husky for this), currently only the coding standards are checked with CI but this could be fine, at least for the time being.

Sephsekla (approved):
This looks good, and we can iterate on it later with different actions etc.

It does touch on quite a few things in different PHP files, so I'd like to keep PRs smaller in future to keep things a bit more modular.

Issue #84, opened 2022-06-13
Demo Feedback: Hub filters


Sephsekla:
As many sites will have a lot of notifications, and there is also the use case of multisite to consider, we should look at adding UI to enable filtering the notification view by priority/source/site.

PR #92, opened 2022-06-14
Notification scripts update

erikyo:
Since the previous version has all hub notifications hardcoded and the hub could not be tested, I created a react component and a redux store controller to handle notifications multiple destinations. The component has a context (formerly called location), which subscribes to notifications from its context and receives updates from it.

Then `notifyController` was rebuilt with all the previous functionality (add, remove, clearAll) and, in addition, a function was added to retrieve the api (it currently reads a false response from a json file).

Actually we have two main components: `Notification` and `NotificationArea`, the latter depending on context displays the array of `Notification` with diffent styles and options (for example at the moment the hub is set to display two lists of notifications, the current week's and previous week notices. [This is not yet customizable](https://github.com/erikyo/wp-notify/blob/443873fc9e93cdbe83e92df64ff1708de85b4b4f/src/scripts/components/notifications.js#L72-L80) but it would be nice if we can provide some filters (by type, by plugin etc as mentioned in PR https://github.com/WordPress/wp-feature-notifications/issues/84)

We have a rest-api endpoint (thanks @Sephsekla!) but we need to decide how the component fetches data from the database, what arguments it passes to it, and the strategy for updating the data in the reducer.

a brief story of what is changed with this PR:
- Adopted the `@wordpress/components` in order to comply to wordpress design standards
- Adopted the `@wordpress/data` to dispatch notifications to all components that request them
- The old function `addNotify` has been added to window.wp.notify (the js/api was located there)
https://github.com/erikyo/wp-notify/blob/443873fc9e93cdbe83e92df64ff1708de85b4b4f/src/scripts/wp-notify.js#L50
- When you test this PR remember to update the timestamp in the json file and this way moment.js will tell you how much time has passed in a textual way ("an hour ago", "few second ago", "a week ago" and so on)
- Adds the storybook and an initial component setup
- Fixes some a11y related issues (mainly now the hub is operable via keyboard)
- UI: the vertical padding of dash notifications should be a little smaller now on horizontal monitors with low vertical resolutions, because now it depends on the height of the screen
- Settings table updated (#110)


includes/demo.php
```diff
@@ -16,11 +16,10 @@ function wp_admin_bar_wp_notify_item( $wp_admin_bar ) {
 	}
 
 	$footer = '<footer>
-    <a>
       <a href="' . esc_url( admin_url( 'options-general.php?page=wp-notify' ) ) . '" class="wp-notification-action wp-notification-action-markread button-link">
```
Sephsekla: Not important since this is just demo code, but in future I'd probably use `printf` for these in order to make things a bit easier to read.

package.json
```diff
@@ -2,17 +2,17 @@
...
-	"homepage": "https://github.com/WordPress/wp-notify#readme",
+  "keywords": [
+    "WordPress",
+    "notifications",
+    "dashboard"
+  ],
+  "author": "The WordPress Contributors",
+  "license": "GPL-2.0-or-later",
+  "bugs": {
+    "url": "https://github.com/WordPress/wp-notify/issues"
```
Sephsekla:
```suggestion
    "url": "https://github.com/WordPress/wp-feature-notifications/issues"
```

While we're in here

package.json
```diff
@@ -2,17 +2,17 @@
...
+  },
+  "homepage": "https://github.com/WordPress/wp-notify#readme",
```
Sephsekla:
```suggestion
  "homepage": "https://github.com/WordPress/wp-feature-notifications#readme",
```

package.json
```diff
@@ -34,22 +34,47 @@
...
     "test:e2e": "wp-scripts test-e2e",
-    "test:unit": "wp-scripts test-unit-js"
+    "test:unit": "wp-scripts test-unit-js",
+    "storybook": "start-storybook -p 6006",
+    "build-storybook": "build-storybook",
+    "docGen": "npx docgen src/scripts/wp-notify.js"
+  },
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/WordPress/wp-notify.git"
```
Sephsekla:
```suggestion
    "url": "git+https://github.com/WordPress/wp-feature-notifications.git"
```

src/scripts/components/noticeImage.js
```diff
@@ -0,0 +1,25 @@
+import { Component } from '@wordpress/element';
+
+import wpLogo from '../../images/WordPressLogo.svg';
+
+export default class NoticeImage extends Component {
```
Sephsekla: I have a few thoughts on this, but most can wait for a followup later.

Each of these components could do with a comment though.

src/scripts/components/notification.js
```diff
@@ -0,0 +1,126 @@
...
+		dismissible: false,
+		onDismiss: () => async () => {
+			delay( 100 ).then( () => this.unmount );
+		},
+	};
+
+	render() {
+		const classes = [
+			'wp-notification',
+			'wp-notice-' + this.props.id,
```
Sephsekla: As a general note, I personally tend to prefer something like

```js
const { additional, dismissible, id /* ..etc */ } = this.props
```

up top, and then you can just refer to `id` or `dismissable` in the rest of the code.

src/scripts/components/notification.js
```diff
@@ -0,0 +1,126 @@
...
+// Register components
+import { Component } from '@wordpress/element';
+import { Button } from '@wordpress/components';
+import NoticeImage from './noticeImage';
+
+import moment from 'moment';
+import { delay } from '../utils';
+
+// Notice class method.
+export default class Notice extends Component {
```
Sephsekla: Does this have to be a class? Or could it just be `const` or `function`? It's not got any class-specific logic or anything.

src/scripts/components/notification.js
```diff
@@ -0,0 +1,126 @@
...
+
+	render() {
+		const classes = [
+			'wp-notification',
+			'wp-notice-' + this.props.id,
+			this.props.dismissible ? 'is-dismissible' : null,
+			this.props.severity || null,
+			this.props.unread ? 'unread' : null,
+			this.props.additionalClassName,
+		].filter( Boolean );
```
Sephsekla: Have you come across https://github.com/JedWatson/classnames? This will do a lot of the heavy lifting here for you.

It's already in use in Gutenberg, so there's a precedent.

src/scripts/components/notification.js
```diff
@@ -0,0 +1,126 @@
...
+
+		return (
```
Sephsekla: This is good, but it is a bit of a massive thing 😅 

Maybe not now, but it would be good to split it up slightly, so something like this for each of the individual conditional components.

```js

const dashboardActions = () => <something>Some buttons</something>

return (
  <div>
    { (location === 'dashboard') && dashboardActions }
  </div>
)

File: src/styles/notify/notification-hub.scss [hunk 10]
```diff
@@ -24,6 +24,11 @@
     transform: translateX(-100%);
   }
 
+  // reset child icons color since by default the icon is lightened when the mouse is in hover state
+  #wpadminbar &:hover #wp-notify-hub .ab-icon:before {
```
Sephsekla: These hover states should probably apply to `:focus` as well, since keyboard nav should have the same experience

src/styles/notify/notification-hub.scss
```diff
@@ -24,6 +24,11 @@
     transform: translateX(-100%);
```
Sephsekla: I'd like to split some of these files up a bit further in an ideal world.

Sephsekla (commented):
@erikyo thanks for all your hard work here! I've added a few comments but in general this has massively pushed things forward. 

As a general note I'd like to do some work on modularisation and comments in particular, but I think it would make more sense to get this merged and then we can chip away at any updates in smaller PRs (since this is pretty huge at this point).

@bacoords @jonathanbossenger what are your thoughts? Should we go ahead with a merge so we can work against `develop` again?

src/scripts/components/notification.js
erikyo: For certain things I have always preferred to create a class whereas I use functional components more often for component elements, this one has also a method (onDismiss) and I guess more may be added in the future..
In this way the component is also more prone to being extended (not that it would be impossible otherwise)

src/scripts/components/notification.js
erikyo: Following this idea I also created a component for the actions (ok+dismiss button), but I was wondering if it would be a good idea to do the same for the title/text. What do you think?

Sephsekla (commented):
👍 We can definitely iterate on this, but for now (once conflicts are resolved) I'd be in favour of getting this merged so we can move to smaller PRs against `develop` (since a lot of the progress is still only in this branch.

erikyo:
since [989c918](https://github.com/WordPress/wp-feature-notifications/pull/92/commits/989c918b912c8acd5634abeb00ffc3274e9eb59f) checks are failing since i've removed the rules exclusions in the phpcs configuration file. 

I think if we want use the WP-standards we can't pick and choose the ones we like or don't like and that since it's only loaded the ruleset of core and doc we should follow it.

Sephsekla:
> since [989c918](https://github.com/WordPress/wp-feature-notifications/pull/92/commits/989c918b912c8acd5634abeb00ffc3274e9eb59f) checks are failing since i've removed the rules exclusions in the phpcs configuration file.
> 
> I think if we want use the WP-standards we can't pick and choose the ones we like or don't like and that since it's only loaded the ruleset of core and doc we should follow it.

@erikyo for now can you just revert the phpcs config change? While I agree that it's a change worth making, I'd prefer to make it in its own followup PR. It's a little outside the scope of JS scripts update, and ideally something to handle separately rather than delaying merging this into develop.

Sephsekla (approved):
💥 Awesome stuff!

Re: the changes to the json file, I think this is fine for now, although I'd like to get the ball rolling again on the schema discussion and finalise this before we go much further.

There's a lot in here and probably a few things that could be tweaked, but for now I'd be in favour of merging this and iterating on things in smaller PRs going forward.

Sephsekla (approved):
🥳

PR #93, opened 2022-06-15
Storybook

erikyo:
## What?
[Storybook ](https://storybook.js.org/) is fantastic way to show React components into a living playground, this pr add the support for that tool

## Why?
It is not necessary to the working plugin but this PR adds a visual documentation about the few components we currently using. To mention, gutenberg has it's own playground done with storybook -> https://wordpress.github.io/gutenberg/?path=/story/docs-introduction--page

## Testing Instructions
to start the local server with the storybook page:
`npm i && npm storybook `

## What is Missing
At the moment was just installed and configured in order to import the single notice.
- we need to create the variants
- wp style isn't fully loaded, seems the font is missing

storybook


Sephsekla:
Tagged myself so I don't forget to look at this. I know it's still a draft but I'd like to try it out.

erikyo:
@Sephsekla this is the [basic notification](https://github.com/WordPress/wp-feature-notifications/blob/05b51af94df1b316ee2a446456a51ec2e783397f/src/stories/Notification.stories.jsx) component page, just copy this file to start another one. https://storybook.js.org/docs/react/writing-stories/args

I guess we are missing the hub notification (and the hub if we like also that), will be almost the same of the dashboard but with a different container.

erikyo:
closed, since I merged this branch with js-dev (it didn't make sense to have two parallel branches, since stories are likely to use the last component from the other repo)

Issue #107, opened 2022-07-28
Accessibility: alt attribute on wp-notification-image


joedolson:
The images associated with notifications should be considered decorative, but need an empty alt attribute to prevent screen readers from attempting to interpret them.

erikyo:
closed with [#92](https://github.com/WordPress/wp-feature-notifications/pull/92)

https://github.com/WordPress/wp-feature-notifications/blob/680be57193c8f7cab73c48507b97696859ca1037/src/scripts/components/NoticeImage.js#L49

PR #112, opened 2022-08-09
107 - Update decorative images with empty alt

Sephsekla:
## What?

Adds an empty alt attribute to decorative images.

## Why?

Fixes Accessibility: alt attribute on wp-notification-image #107

## How?

Updates demo notification React with additional attribute.


erikyo:
sorry but the correct branch to work is the one named [js-develop](https://github.com/erikyo/wp-notify/tree/js-develop) and the fix was already applied... in addition [that part with the hardcoded hub](https://github.com/WordPress/wp-feature-notifications/blob/0fabf6a818240bba8e2d937d8c4b62c1b99527f3/src/scripts/wp-notify.js) was replaced with the fake-api.js so i guess that will be lost at the first merge

Sephsekla:
Oh oops, felt like doing something quick, hadn't realised that was the case.

I'll close this then, do you mind updating the issue so we're clear on what's already being worked on?

erikyo:
Yep sorry about this! Maybe this is one of those things that we should point out to the other folks in the group during meetings...

PR #117, opened 2022-10-26
Add demo REST controller

Sephsekla:
## What?

This PR adds a demo REST Controller for fetching notifications.

## Why?

This allows us to move the demo content to the PHP side, allowing real js querying of the endpoint.

### DIFF
--- a/includes/restapi/class-notification-controller.php
+++ b/includes/restapi/class-notification-controller.php
@@ -1,0 +1,86 @@
+<?php
+/**
+ * REST API controller to get and set notifications.
+ *
+ * @package wordpress/wp-feature-notifications
+ */
+
+namespace WP\Notifications\REST;
+
+use WP_Error;
+use WP_REST_Controller;
+use WP_REST_Request;
+use WP_REST_Response;
+use WP_REST_Server;
+
+/**
+ * REST API Controller class
+ */
+class Notification_Controller extends WP_REST_Controller {
+
+	/**
+	 * Namespace for the REST endpoint.
+	 *
+	 * @type string
+	 */
+	const NAMESPACE = 'wp/v2';
+
+	/**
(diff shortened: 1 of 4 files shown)


includes/restapi/class-notification-controller.php
```diff
@@ -0,0 +1,84 @@
...
+		register_rest_route(
+			self::NAMESPACE,
+			self::NOTIFICATION_BASE,
+			array(
+				'methods'             => WP_REST_Server::READABLE,
+				'args'                => array(),
+				'callback'            => array( $this, 'get_notifications' ),
+				'permission_callback' => array( $this, 'authenticate' ),
+			),
+			false,
```
erikyo:
```suggestion
			false
```
@Sephsekla if you could remove this comma that causes the failing action, we can merge this commit, so I start the rebase with the js branch with also this commit

Sephsekla:
Hmm, looks like an issue with PHPCS now, don't think it's related to my code but I'll take a look later.

erikyo:
I think we need to add 
`excludes: 'tests/'`

into actions WPCS action template. (check [here](https://github.com/erikyo/cf7-antispam/blob/main/.github/workflows/wpcs.yml))

Sephsekla:
> I think we need to add `excludes: 'tests/'`
> 
> into actions WPCS action template. (check [here](https://github.com/erikyo/cf7-antispam/blob/main/.github/workflows/wpcs.yml))

Hmm, weird that this test passed on the previous commit though.

erikyo:
indeed i was wrong, it is not an issue of with excluded paths and looking closely even files inside "./includes/" have the same problem. 

if you have the time try [updating wpcs](https://github.com/10up/wpcs-action/tree/stable), the one we are using is an old version

erikyo (approved):

Issue #110, opened 2022-07-28
Accessibility: settings checkbox labels are not unique


joedolson:
### Description

Each notification source 's collection of checkboxes has the same ID. IDs should be unique, with unique labels. With these relationships, the accessible name produced for any given will be unpredictable, depending on how that browser selects the labeling value when there are multiple IDs with different values. (In Chrome, for example, the 'SMS' checkbox comes up with no accessible name.)

In the current model, these settings are difficult to parse non-visually, since they have three separate parts (table row heading, table column heading, 'Show In' instructive text, and the label itself. 

I'd suggest remodeling these settings using fieldset/legend relationships so that the context of each grouping is more manageable. 

At minimum, the label values need to be unique; perhaps using `aria-describedby` to associate with the visible headings instead of using a visually hidden `label` element. 

### Step-by-step reproduction instructions

1. Inspect code. 
2. Observe the duplicated IDs.

### Screenshots, screen recording, code snippet

![image](https://user-images.githubusercontent.com/1686451/181610503-1b08cfa2-6807-4a1f-a876-97e771ba76d0.png)

### Please confirm that you have searched existing issues in the repo.

Yes

### Please confirm that you have tested with all plugins deactivated except WP Feature Notifications.

Yes

erikyo:
This one is related to a [commit ](https://github.com/WordPress/wp-feature-notifications/pull/92/commits/b8d77957d0ee9ea190ef9bb1a93a0c46ced506e2) that has been reverted because it didn't address the scope of #92 (furthermore coding standards needed to be fixed). I reopen because it will be moved into a dedicated PR

erikyo:
closed with #145

https://github.com/WordPress/wp-feature-notifications/pull/145/files#diff-c8ae8b0d03b0f11bcf4f7a601722e23bdacc8bf0ebfa49bd6a1fbbb89c245864R180-R188

PR #145, opened 2023-01-20
Updated settings demo page

erikyo:
## What?
Updates the settings page demo code

## Why?
Accessibility: settings checkbox labels are not unique #110

## How?
The settings table has been refactored, now it is built by a couple of loops and the ids are dynamically generated so there should be no more duplicate ones

## Testing Instructions
- Start wp-env 
- navigate to http://localhost:8888/wp-admin/options-general.php?page=wp-notify


includes/class-wp-notify-demo.php
```diff
@@ -0,0 +1,249 @@
+<?php // phpcs:ignore WordPress.Files.FileName.InvalidClassFileName
```
Sephsekla: Can I ask why this file has been renamed?

It's functional code rather than a class, so feels inaccurate,

includes/class-wp-notify-demo.php
```diff
@@ -0,0 +1,249 @@
...
+/**
+ * @package wp-feature-notifications
+ *
+ * Development demo files
+ **/
+
```
Sephsekla: It's something I'm planning on taking on as a larger project, but what do you think about using a namespace for this instead of prefixes? 
Probably

```php
namespace WP\Notifications\Demo;
use WP_Admin_Bar;
```
etc

includes/class-wp-notify-demo.php
erikyo: Let's say it was the beginning of trasposing "demo" to be a real plugin file, imho it wouldn't hurt to create an init class for the notices plugin. Moreover, actually there is a class for tables inside this file should be isolated.
Do you think it would be better to do it directly in this PR or leave it for when this is accepted? frankly, it would be better if you did this one!

Sephsekla: I think having it in the class file without fully restructuring is going to cause some problems later tbh. What do you think about renaming it back to demo.php for now, and then I can pick up refactoring the PHP once this is merged if that works for you?

bacoords:
This looks good.  A few random thoughts here:

- We'll need a way for plugins/themes to register or hook into this table to add their own rows- perhaps based on a notification source/category or else
- I sure would love to see an example of a new WP settings page built with JS components instead of WP List Table (even though the WP List table here is a big improvement over the previous static HTML). Is there anything in core using JS for admin/settings areas outside the block/site editor?
- This screen is missing some sort of "save" button

includes/demo.php
```diff
@@ -84,173 +84,129 @@ function wp_notify_add_admin_options_page() {
...
+	/**
+	 * The settings table
+	 *
+	 * @return void
+	 */
+	function init_table() {
+		/** @var $list_table $list_table */
+		$list_table = new list_table();
+		$list_table->prepare_items();
+		$list_table->display();
```
erikyo, line 219:
```suggestion
		$list_table->display();
		echo '<input type="submit" id="doaction" class="button action primary" value="apply">';
```

will add the missing confirmation button

erikyo:
1 - About the first thing (giving plugins the possibility of adding fields) I had thought about it. We just add a filter [here to add a row](https://github.com/WordPress/wp-feature-notifications/pull/145/files#diff-c8ae8b0d03b0f11bcf4f7a601722e23bdacc8bf0ebfa49bd6a1fbbb89c245864R157-R162) but it is [a bit more complicated to add a column](https://github.com/WordPress/wp-feature-notifications/pull/145/files#diff-c8ae8b0d03b0f11bcf4f7a601722e23bdacc8bf0ebfa49bd6a1fbbb89c245864R168-R176)... it is not impossible but at the moment I did not want to create much code for it as we still have to define it well (see next point)  

2 - As a great supporter of "js for everything" I would say no because it would be completely different to the standard WordPress "way", and besides I don't see the need. Anyway this is a possibility I have explored already [in this repo](https://github.com/erikyo/plugin-ui/blob/main/src/scripts/components/SettingsForm.js) where I made a settings panel that saves using the rest-api. I thing that could be nice for a plugin with a lot of settings (like woocommerce) but in our case could be an unnecessary overhead. 

3 - 🤦the table does not provide a confirmation button because it is not a form. this could [fix](https://github.com/WordPress/wp-feature-notifications/pull/145/files#r1111274519)

includes/demo.php
```diff
@@ -84,173 +84,129 @@ function wp_notify_add_admin_options_page() {
...
+
+		private function table_data(): array {
+			$data = array();
+
+			$indexes = array(
+				'WordPress',
+				'User activity (published, edited, ect)',
+				'Comments',
+				'Site health',
+			);
```
erikyo: With this change we can add items to 'who' emits the messages

```suggestion
			/**
			 * Notification sources
			 *
			 * Allows to filter the notification sources that will be shown in the table.
			 *
			 * @since 0.0.2
			 *
			 * @param array $indexes Array of available notification emitters
			 */
			$indexes = apply_filters(
				'wp_feature_notifications_settings_indexes',
				array(
					'WordPress',
					'User activity (published, edited, ect)',
					'Comments',
					'Site health',
				)
			);
```

erikyo: > We'll need a way for plugins/themes to register or hook into this table to add their own rows- perhaps based on a notification source/category or else

In the same way we can provide a filter to expand or remove the 'way' you receive notifications. But perhaps for a slightly more complex operation this approach is not good enough and we will have to register the plugins that use the wp-notify. What do you think? Could this at least temporally provide a solution to the issue?

includes/demo.php
Sephsekla: If you're going to add this, could you make the `apply` uppercase and translatable?

erikyo:
```suggestion
		$list_table->display();
		printf( '<input type="submit" id="doaction" class="button action primary" value="%s">', __( 'Apply' ) );
```

right... I would say it is now OK, only the text domain is missing but at least on this page it has never been used yet

includes/demo.php
Sephsekla: I think this works well, at least in the short term. Longer term I think we'll have something more advanced set up.

Sephsekla:
I think your suggestions make sense. Could you update this with the latest `develop` branch and merge your changes? We can iterate on it later but for now I'd like to get it done so we can unblock the e2e tests.

includes/demo.php
erikyo: However, I should probably also add the filter docblock. correct?

Sephsekla: Ah very good point, I always forget that for filters.

erikyo: updated!

Sephsekla (approved):

PR #146, opened 2023-01-20
hides /src/stories/assets and package.json from linguist

### LINKED ISSUE #138: Add the .gitattributes in order to hide some folders/files from linguist
Currently are counted by linguist files that are not really part of this repo, for example the storybook assets folder which uses all the WordPress backend styles or package.json which is a generated file

add a .gitattributes with an exclusion like the following:
```text
# remove generated / documentation 
package.json linguist-generated=true
/src/stories/** linguist-vendored=true
```

erikyo:
## What?
Exclude vendored and generated files

## Why?
https://github.com/WordPress/wp-feature-notifications/issues/138

## How?
added to .gitattributes this content
```
# exclude vendored and generated files
package.json linguist-generated=true
/src/stories/assets/** linguist-vendored=true
```

close https://github.com/WordPress/wp-feature-notifications/issues/138

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,4 +1,4 @@
 # exclude vendored and generated files
-package.json linguist-generated
+package-lock.json linguist-generated
 /src/stories/assets/** linguist-vendored
 /docs/** linguist-documentation
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,3 @@
+# exclude vendored and generated files
+package.json linguist-generated=true
+/src/stories/assets/** linguist-vendored=true
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,3 +1,4 @@
 # exclude vendored and generated files
-package.json linguist-generated=true
-/src/stories/assets/** linguist-vendored=true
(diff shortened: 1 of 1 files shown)


.gitattributes
```diff
@@ -0,0 +1,4 @@
+# exclude vendored and generated files
+package.json linguist-generated
+/src/stories/assets/** linguist-vendored
+/docs/** linguist-generated
```
Sephsekla: Not that familiar with this, but looking at https://github.com/github/linguist/blob/master/docs/overrides.md:

```suggestion
/docs/** linguist-documentation
```

Sephsekla (commented):
How about adding package-lock.json as well?

.gitattributes
erikyo: yep looks like linguist-documentation and lingust-generated are aliases. documentation seems more appropriate in any case so I'll accept this suggestion

Sephsekla (dismissed):
Any thoughts on the package-lock suggestion? Good to go either way.

Sephsekla (approved):

Issue #154, opened 2023-02-13
Add e2e tests


erikyo:
## What problem does this address?
it would be better to automate e2e's tests in order to check in advance whether some update or contribution breaks the plugin 

## What is your proposed solution?
Create some e2e tests to check with jest that everything is working properly

erikyo:
closed with https://github.com/WordPress/wp-feature-notifications/pull/155

PR #155, opened 2023-02-13
e2e testing + nodejs ci

erikyo:
## What?
this PR is for adding e2e tests to wp-notify

## Why?
As discussed recently on slack, we think it may be convenient to check beforehand the correctness of contributions and, if we decide to activate the auto-update of depend bots, to check that suggested packages do not break the plugin

#154 - Add e2e tests

## How?
I added jest and puppeteer and with these I created (thanks to @wordpress/e2e-tests) two 'projects', one for unit tests and the other for e2e tests (see the jest.config.js file for more info). Then I added in package.json the 'test' script that simply launches jest (both projects)
As far as CI is concerned, I moved what was related to js to 'node-ci.yml' and then i added the e2e test job after the build testing


Sephsekla:
Blocked until https://github.com/WordPress/wp-feature-notifications/pull/145 is merged

Sephsekla:
@erikyo hmm, looks like npm install is failing on build, could you take a look?

Sephsekla:
@erikyo this is still failing on node 14. However I'm wondering if we should just drop this test altogether tbh. Node 14 goes out of support next month, and it's not like PHP where we should be prepared to run it on different end user environments, particularly with a .nvmrc file. What do you think?

erikyo:
I think isn't critical, but on the other hand I wanted to stay in line with Gutenberg which is on node 14 yet. 

Unfortunately I haven't had time to look at this PR last days but I believe that a good solution for the future would be to reorganise the dependencies better, perhaps by moving the storybook stuff into a separate package.json (inside the stories folder). 

Maybe we could add the build test for different OS that can be more useful compared to an old node version! 

@johnhooks, you are too of the idea of removing node 14, is that right?

johnhooks:
@erikyo yes I think I told you something almost identical to @Sephsekla points 😆 in a DM. Though I agree if Gutenberg doesn't plan to bump their required Node version, it would be best to figure it out. Is anyone knowledgeable about the status of bumping to 16?

johnhooks:
Looks like it's in the works, it might even get bumped to 18 🎉. I would drop it and we can revisit the decision when/if it ever becomes a problem.

https://github.com/WordPress/gutenberg/issues/48588

Sephsekla:
OK, looking through the WP Core PR I'd agree. A little tempted to upgrade the nvm version to 18 as well. @erikyo if you can update the PR with the latest branch then I can approve.

Sephsekla (approved):
💥

Issue #166, opened 2023-03-16
The file .nvmrc has a wrong encoding


erikyo:
### Description

It has been reported to us by @johnhooks that (at least on mac) the `nvm use` command returns an error and on closer inspection it is due to the fact that it uses an 'UTF-16LE' encoding 

### Step-by-step reproduction instructions

check this file:
https://github.com/WordPress/wp-feature-notifications/blob/680be57193c8f7cab73c48507b97696859ca1037/.nvmrc#L1

### Screenshots, screen recording, code snippet

![image](https://user-images.githubusercontent.com/8550908/225613956-a7dd2a13-eccc-4374-8f44-2fbf17c6123e.png)

### Environment info

not applicable

### Please confirm that you have searched existing issues in the repo.

Yes

### Please confirm that you have tested with all plugins deactivated except WP Feature Notifications.

Yes

erikyo:
closed with https://github.com/WordPress/wp-feature-notifications/pull/169

PR #169, opened 2023-03-20
Fix .nvmrc file

Sephsekla:
## What?

Recreate .nvmrc file due to encoding issues. I've gone with lts/gallium for now, as the build seems to work fine with this, however we can revisit the exact version down the line.

## Why?

See #166 

## How?

- Delete old .nvmrc file entirely
- Create new .nvmrc file

## Testing Instructions

- Use nvm to set correct node version
- Check for successful build


erikyo (commented):
Don't you think it would be better to use the LTS as a target? In this way we can also avoid updating this file

Sephsekla:
> Don't you think it would be better to use the LTS as a target? In this way we can also avoid updating this file

Hmm, it's a fair point. On the other hand I'd argue that switching the node version should be something that we deliberately do when the time comes.

erikyo (approved):
LGTM!

Issue #176, opened 2023-03-24
Add rest-api endpoint to dismiss a notification


johnhooks:
## What problem does this address?

The current rest-api does not support dismissing a notification.

## What is your proposed solution?

I propose adding and a new endpoint:

```
POST /wp-json/wp-notifications/v1/notifications/[NOTIFICATION_ID]
```

To dismiss the message, the new status should be in the request body, something like:

```json
{
  "status": "dismissed"
}
```

A combination of the user ID the message ID and the post body will provided all the necessary information to update the message's status in the database.

PR #178, opened 2023-03-24
feature(mvp): Stage 4 add item and collection schemas to rest api

### LINKED ISSUE #194: Change rest-api namespace to `wp-notifications/v1`

The plugin should not use the core `wp/v2` namespace 

@TimothyBJacobs suggested `wp-notifications/v1` in #179, which sounds good.

Another thought is to stick with the "Notify" name and do `wp-notify/v1`, this is shorter, and if this does merge into core, the "wp-notify" namespace/prefix used everywhere would simply become "wp".

johnhooks:
## What?

Implement item and collection schemas for notification the rest-api. Initiate PR early to get feedback, still lots to figure out.

Starts work on #176 to create an update rest endpoint for notifications.

Fixes #194

## Why?

The rest-api of notifications should implement item and collection schemas.

## How?

It adds schemas that the `WP_Rest_Controller` can use to validate api requests.

## Testing Instructions

Will need to do more research, but I'll be researching core tests and implement something similar.


includes/restapi/class-notification-controller.php
```diff
@@ -41,27 +42,36 @@ public function __construct() {
...
-	 * Authenticate notification request.
-	 *
-	 * @return boolean true if authenticated, false otherwise.
+	 * Checks if a given request has access to read notifications.
+	 * @param WP_REST_Request $request Full details about the request.
+	 * @return true|WP_Error True if the request has access to view the items, error object otherwise.
 	 */
-	public function authenticate() : bool {
-		// TODO: Implement authentication.
+	public function get_items_permissions_check( $request ) {
```
johnhooks, line 68: @Sephsekla This and `get_notifications` where renamed, because `get_items` and `get_items_permission_check` are standard names used in core rest-api.

includes/restapi/class-notification-controller.php
```diff
@@ -81,4 +91,144 @@ public function get_notifications( $request ) {
...
+			'type'        => 'string',
+			'enum'        => array(
+				'undisplayed',
+				'displayed',
+				'dismissed',
+			),
+		);
+
+		return $query_params;
+	}
```
johnhooks: These were just some that I thought might be helpful, please help with determining what would be best.

includes/restapi/class-notification-controller.php
```diff
@@ -81,4 +91,144 @@ public function get_notifications( $request ) {
...
+					'type'        => 'string',
+					'context'     => array( 'view', 'edit', 'embed' ),
+				),
+				'meta'         => array(
+					'description' => __( 'The metadata of the notification.' ),
+					'type'        => 'string',
+					'context'     => array( 'view', 'edit', 'embed' ),
+				),
+			),
+		);
```
johnhooks: This implements the database schema I suggested in #168, I can walk this back to the original schema if that suits everyone better.

johnhooks: I added "context" in as suggested by @erikyo

includes/restapi/class-notification-controller.php
```diff
@@ -81,4 +91,180 @@ public function get_notifications( $request ) {
...
+				'status'       => array(
+					'description' => __( 'The status of the notification.' ),
+					'type'        => 'string',
+					'default'     => 'undisplayed',
+					'enum'        => array(
+						'undisplayed',
+						'displayed',
+						'dismissed',
+					),
+				),
```
johnhooks: @erikyo how do these look to you?

- `context` defaults to `hub`
- `severity` is nullable

Does that match the usage?

erikyo: to me they are fine but it seems to me that it is missing `info` that is currently used with [WordPress](https://developer.wordpress.org/reference/hooks/admin_notices/)

includes/restapi/class-notification-controller.php
erikyo: The only one that seems missing (to me) is the "count" to limit the number of queried objects

johnhooks: That is handled by the `WP_REST_Controller` that this class inherits from, and it's called "page". But if I'm wrong, and we have to add it ourselves, I'll take a look at the wp-json standard request parameters for limited the number of items and use the same.

erikyo: In this case I would say everything is there, and if something is missing, we can add it as we go!

@Sephsekla what do you think? Do you think the possible parameters are all there, at least the main ones for an MVP?

johnhooks: @erikyo, the page parameters are handled by the parent rest controller method `get_collection_params` https://github.com/WordPress/wordpress-develop/blob/trunk/src/wp-includes/rest-api/endpoints/class-wp-rest-controller.php

johnhooks:
@Sephsekla this is ready to review. 

Note: I'm not trying to commit us to anything, but just start work toward an actual rest-api. If anything in this needs to be adjusted as we work to implement notifications, I am 100% on board to either make the changes or support anyone doing the work.

tests/phpunit/tests/rest-api/test-notification-controller.php
```diff
@@ -0,0 +1,81 @@
...
+		$this->markTestSkipped( 'TODO Implement' );
+	}
+
+	public function test_delete_item() {
+		$this->markTestSkipped( 'TODO Implement' );
+	}
+
+	public function test_prepare_item() {
+		$this->markTestSkipped( 'TODO Implement' );
+	}
```
johnhooks, line 36: @Sephsekla These are required by the rest test interface. We will fill out this test suite as we build out the rest-api.

includes/restapi/class-notification-controller.php
Sephsekla: This looks good to me, the more we can just use from the base controller, the better. We can always extend it, as well.

tests/phpunit/tests/rest-api/test-notification-controller.php
Sephsekla: Good call. I'd like to make sure we add each test at the same time as its associated functionality.

includes/restapi/class-notification-controller.php
```diff
@@ -72,7 +82,7 @@ public function authenticate() : bool {
 	 *
 	 * @return WP_REST_RESPONSE|WP_Error REST response or WP Error
 	 */
-	public function get_notifications( $request ) {
+	public function get_items( $request ) {
```
Sephsekla, line 86: Hmm, do you reckon we can add an argument type? I know we can't do a union type for the return value on our targeted PHP versions, but maybe

```php
public function get_items( WP_REST_Request $request ) {
```

tests/phpunit/tests/rest-api/test-notification-controller.php
```diff
@@ -0,0 +1,81 @@
+<?php
+/**
+ * Unit tests covering WP_REST_Notification_Controller functionality.
+ */
+class WP_Test_REST_Notification_Controller extends WP_Test_REST_Post_Type_Controller_Testcase {
```
Sephsekla: Do we want to extend the post type controller test case here? I'd have thought we'd use `WP_Test_REST_Controller` as a starting point, given we haven't used the post controller for the notifications controller itself.

johnhooks: @Sephsekla that was a mistake. I was using the Post controller test as a reference (the name is so long), I didn't catch the Post in there. I will change it.

includes/restapi/class-notification-controller.php
johnhooks: My logic was to follow core, and the comment is how they add types. To me it is odd if we have half the type in the comment and half in the signature. But I can add it in.

tests/phpunit/tests/rest-api/test-notification-controller.php
johnhooks: @Sephsekla I totally agree. That is why I added tests for the schema. We could easily get behind on testing if we don't stay on top of it.

tests/phpunit/tests/rest-api/test-notification-controller.php
johnhooks: @Sephsekla should I use namespaces in the tests or name test classes with a prefix?

includes/restapi/class-notification-controller.php
johnhooks: @Sephsekla that actually breaks the typing in my editor (VSCode), because it doesn't pick up types from a mix of comments and function signature. If I add `WP_REST_Request` it wants the return type to be in the function signature too, if it's added it breaks linting because unions are not supported.

The type hints still work, even though in the comment.

tests/phpunit/tests/rest-api/test-notification-controller.php
johnhooks: The test class to inherit from is actually `WP_Test_REST_Controller_Testcase`? `WP_Test_REST_Controller` doesn't exist... or at least I get warnings about it in the IDE.

johnhooks:

erikyo (approved):
LGTM 🚀

PR #181, opened 2023-03-25
build: add jsconfig.json and @types/jest

johnhooks:
This fixes Intellisense auto complete in tests, and fixes the `eslint` errors in the test files.

## What?

Add `jsconfig.json` file and `@types/jest` as a dependency to improve IDE experience in tests. And fix `eslint` errors in tests of `describe` and `it` being `undefined`.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -5,6 +5,7 @@ module.exports = {
 	env: {
 		browser: true,
 		es2021: true,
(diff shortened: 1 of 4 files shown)


erikyo (approved):
LGTM, a better setup is certainly a welcome addition! thanks @johnhooks

PR #184, opened 2023-03-25
build: add @wordpress/env as dev dep

johnhooks:
## What?

Add `@wordpress/env` as a development dependency and add `wp-env` scripts.

## Why?

If using `nvm` to manage projects with different node versions, it's best to keep development dependencies in the `package.json` and installed in a project. Otherwise it has to be reinstalled globally in multiple node version.


erikyo (dismissed):
LGTM

erikyo (approved):

PR #199, opened 2023-03-31
feature: add activator and uninstaller

johnhooks:
## What?

Adds plugin activator and uninstaller. The database SQL to create the database tables for the plugin are in the `Activator::create_tables_v1` static method.

## Why?

Need to be able to install the plugin and have a database to contain the notification data.

## TODO

- [x] register uninstall
- [x] multisite support
- [x] basic unit tests
- [x] research [maybe_create_table](https://github.com/johnhooks/wordpress-develop/blob/a627e1bf21c27cdd816f58df3bc23177de32f760/src/wp-admin/includes/upgrade.php#L2464) as suggested by @Sephsekla, the implementation could be helpful for unit testing activation.
- [x] ~multisite unit tests~ multisite functionality needs more thought before writing tests


erikyo:
In this way, however, it would not work for multisites... Since the code for adding support is not that complex,  why we don't add it directly from the beginning? What do you think @johnhooks, multisite support is not essential but certainly welcome

johnhooks:
@erikyo do you have a concept of how best to make this multisite compatible?

erikyo:
something like this I guess it could fit
https://github.com/erikyo/cf7-antispam/blob/1047eb25ec63c68198696bcf0baaaa5e512ea9e8/includes/cf7a-activator.php#L264-L278

Additional info here:
https://wisdmlabs.com/blog/build-multisite-compatible-wordpress-plugin/

johnhooks:
@erikyo it looks like the real magic is `switch_to_blog( $blog_id )`.  Which I assume modifies `$wpdb->prefix`.

https://github.com/erikyo/cf7-antispam/blob/1047eb25ec63c68198696bcf0baaaa5e512ea9e8/includes/cf7a-activator.php#L271

Though in this instance, we should alway create a set of tables for the main site as well, correct?

erikyo:
> Which I assume modifies $wpdb->prefix

yes exactly! Anyway I don't think it's necessary but I would, because in that way we can create a network related set of notifications

includes/class-activator.php
```diff
@@ -0,0 +1,195 @@
...
+		// Engage multisite if in the middle of turning it on from network.php.
+		$is_multisite = is_multisite() || ( defined( 'WP_INSTALLING_NETWORK' ) && WP_INSTALLING_NETWORK );
+
+		if ( $is_multisite ) {
+			// Get all blogs in the network and uninstall the plugin on each one.
+			$blog_ids = $wpdb->get_col( "SELECT blog_id FROM $wpdb->blogs" );
+			// Loop over the individual sites and create tables for each.
+			foreach ( $blog_ids as $blog_id ) {
+				switch_to_blog( $blog_id );
+				self::create_tables_v1();
```
johnhooks: @erikyo should the options be created for every set of multisite tables?

erikyo: In my opinion yes, because every site has its own

johnhooks (commented):
@erikyo and @Sephsekla I think this one is ready for review. We still need to figure out a testing strategy, though it may take some further research.

johnhooks:

Issue #197, opened 2023-03-31
Create naming convention


johnhooks:
## What problem does this address?

Reduce developer confusion.

## What is your proposed solution?

We should decided on a naming convention, document it, and clean up all the old names in the project.

I do not care what the naming convention is, but I began work on new features, and the naming issues in the project make me very uncertain how I should name things. It's also very odd when they have to be used together.

This is related to #4, #190 and #194.

## Convention Todo List

- [x] Plugn name

  - [x] `WP Feature Notifications`
  - ~[ ] `WP Notifications`~
 
- [x] Package name

  - ~[ ] `WP_Feature_Notifications`~
  - ~[ ] `WP_Notifications`~
  - [x] `wordpress/wp-feature-notifications` in [class-notification-controller.php](https://github.com/WordPress/wp-feature-notifications/blob/c8714bda9e65512730c17d431d2dc96f051cd1fd/includes/restapi/class-notification-controller.php#L5)
 
- [x] Text Domain

  The text domain must match the slug of the plugin, see [Text Domain](https://developer.wordpress.org/plugins/internationalization/how-to-internationalize-your-plugin/#text-domains).

  - [x] `wp-feature-notifications`
  - ~[ ] `wp-notifications`~

- [x] Namespace

  The rest controller uses the namespace `WP\Notifications`

  - [x] Should we convert the whole project?
  - ~[ ] Should we use a PSR-4 compliant file structure?~
 
- [x] Class and Interface

  - [x] If using namespaces, should all the `WP_Notify` prefixes be removed?

- [x] Class and Interface files
  
  The rest controller uses the file name `class-*`, no prefix
  
  The rest use `class-wp-notify-*`

  - [x] If converting to namespaces, should all the files be renamed to remove the `wp-notify` prefix?

- [x] Global define variables
  
  The current prefix is `WP_NOTIFICATION_CENTER_*`, found in `wp-feature-notifications.php`

  - [x] Should we use `WP_FEATURE_NOTIFICATIONS_*`
  - ~[ ] Use namespaced const variables~

@Sephsekla, @erikyo pointed out this doesn't consider updates in the JavaScript or CSS.

- [x] CSS Custom Properties

   I suggest using shorter a condensed namespace, the chances are very low that there will be a conflict.
  - [x] Should we use `wp-notifications-*` for class names?
  - [x] Should we use `--wp-notifications-*` for custom properties?

johnhooks:
[...]

PR #201, opened 2023-04-01
feature: add PSR-4 autoloading

johnhooks:
## What?

Add PSR-4 autoloading. Begins work on #4.

## How?

Modifies the `composer.json` with autoloading paths and namespaces the require `vendor/autoload.php` in `wp-feature-notifications.php`

Blocked until #197 is resolved.


johnhooks:
@erikyo does part of the vendor folder have to be checked in to do this?

johnhooks:
This will require a change to the JS CI workflow. There needs to be a composer install step to generate the autoloading functions. We can either add composer to the JS workflow or use a build asset of the plugin zip in the JS environment, which
would include the vendor folder and autoloading files.

erikyo:
https://github.com/WordPress/wp-feature-notifications/actions/runs/4585697787/jobs/8097988162?pr=201

☝️ This is strange since husky has to perform [repo wide php linting ](https://github.com/WordPress/wp-feature-notifications/blob/b09e7b34d83b213a10b1c6d871761257e36dd26d/.husky/pre-commit#L4)

johnhooks:
@erikyo it was because the prepare script wasn't added yet and I hadn't run "husky install".

johnhooks:
Don't filenames have to match the class names for PSR-4 autoloading to work?

https://www.php-fig.org/psr/psr-4/#3-examples

@Sephsekla are we willing to change this lint rule so we can use PSR-4 autoloading?

johnhooks:
Not implementing PSR-4 autoloading to maintain WP coding standards.

PR #202, opened 2023-04-01
Ci / Split tasks to avoid testing issues

erikyo:
## What?
We update ci to separate build tasks with test tasks

## Why?
With the new setup we need composer to build the autoload.php file before testing, otherwise the autoloaded files are missing

## How?
Three files representing the three tasks were created, and in the testing one all the tests were merged
(bonus) added the os testing for node
removed -> php 7.3

### DIFF
--- a/.github/workflows/node-build.yml
+++ b/.github/workflows/node-build.yml
@@ -22,6 +22,6 @@ jobs:
         with:
           node-version: ${{ matrix.node-version }}
           cache: 'npm'
-      - run: npm install
+      - run: npm ci
       - run: npm run build --if-present
 
--- a/.github/workflows/node-build.yml
+++ b/.github/workflows/node-build.yml
@@ -8,11 +8,13 @@ on:
 jobs:
   build:
     name: Build plugin
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        node-version: [16.x, 18.x]
         # See supported Node.js release schedule at https://nodejs.org/en/about/releases/
+        node-version: [16.x, 18.x]
+        os: [linux, macos, windows]
+
     steps:
       - uses: actions/checkout@v3
       - name: Use Node.js ${{ matrix.node-version }}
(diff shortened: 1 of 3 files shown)


.github/workflows/test-suite.yml
```diff
@@ -53,20 +53,23 @@ jobs:
     - name: Install dependencies
       run: composer install --prefer-dist --no-progress
 
+    - name: Setup environment to use the desired version of NodeJS
+      uses: actions/setup-node@v3
+
+    - name: Installing NPM dependencies
+      run: npm install
```
johnhooks: This will use the lock file and throw an error if it needs to be updated by a PR.

```suggestion
      run: npm ci
```

.github/workflows/node-build.yml
```diff
@@ -23,25 +25,3 @@ jobs:
       - run: npm install
```
johnhooks:
```suggestion
      - run: npm ci
```

.github/workflows/test-suite.yml
```diff
@@ -53,20 +53,23 @@ jobs:
...
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/checkout@v3
-      - name: WPCS check
-        uses: 10up/wpcs-action@stable
-        with:
-          enable_warnings: true
-          use_local_config: true
+    - name: Starting the WordPress Environment
+      run: wp-env start
```
johnhooks:
```suggestion
      run: npm run wp-env:start
```

.github/workflows/test-suite.yml
```diff
@@ -53,20 +53,23 @@ jobs:
...
       run: composer install --prefer-dist --no-progress
 
+    - name: Setup environment to use the desired version of NodeJS
+      uses: actions/setup-node@v3
+
+    - name: Installing NPM dependencies
+      run: npm install
+
+    - name: Installing wp-env
+      run: npm -g i @wordpress/env
```
johnhooks, line 63: This shouldn't be necessary because `@wordpress/env` is a dev dependency now.

johnhooks (changes requested):
@erikyo do you think a workflow to run `npm run lint:js` is a good idea?

erikyo:
> @erikyo do you think a workflow to run npm run lint:js is a good idea?

the linting is something i prefer to fire before the ci, for example with a githook before commit (with husky that is already in place)

johnhooks (approved):

.github/workflows/test-suite.yml
```diff
@@ -53,20 +53,23 @@ jobs:
...
       run: composer install --prefer-dist --no-progress
 
+    - name: Setup environment to use the desired version of NodeJS
+      uses: actions/setup-node@v3
+
+    - name: Installing NPM dependencies
+      run: npm ci
+
+    - name: Installing wp-env
+      run: npm -g i @wordpress/env
```
johnhooks: @erikyo let's try this again.
```suggestion
```

johnhooks (changes requested):

johnhooks (approved):
LGTM!

johnhooks (approved):

johnhooks:
PR #213 proposes the same functionality, but extracts the environment setup into custom actions.

PR #204, opened 2023-04-01
feature: add basic Logger class

johnhooks:
## What?

Add basic Logger class.

## Why?

There isn't currently a good solution for logging debug information.

Blocked until https://github.com/WordPress/wp-feature-notifications/issues/197 is resolved.


johnhooks:

PR #207, opened 2023-04-01
build: fix husky pre-commit hook

### LINKED ISSUE #206: `husky` pre-commit hook is not executable
On git commit I get an error. 

```
hint: The '.husky/pre-commit' hook was ignored because it's not set as executable.
```

I believe `chmod +x .husky` should fix the issue.

johnhooks:
## What?

Make the husky pre-commit hook executable.

Fixes #206

## Why?

The pre-commit script was being ignored because it was not executable.

## How?

`chmod +x .husky`


erikyo (approved):
LGTM

PR #211, opened 2023-04-02
build: add lint-staged

johnhooks:
## What?

Add and config `lint-staged`.

## Why?

Currently the whole `includes` folder is linted in the pre-commit hook, which prevents commits when there were previous issues. Also the `wp-feature-notification.php` and the php tests are currently not being linted.

Adding lint-staged will allow linting only files that are staged for the commit


erikyo (approved):
image

@johnhooks good job! it works perfectly thanks!

PR #217, opened 2023-04-03
build: fix lint-staged config

johnhooks:
## What?

Update the lint-staged config.

## Why?

To fix an error @erikyo found with using an older configuration style.


erikyo (approved):
LGTM 🎉

PR #213, opened 2023-04-02
ci: refactor github workflows

johnhooks:
## What?

Refactor the GitHub CI workflows.

Abstract out Node.js and PHP environment setup into actions.

## Why?

The current CI workflows do not properly setup the PHP environment for JS e2e testing, if the project implements PSR-4 autoloading. Composer needs to run to generate the autoloading files.

Prepares for #201.

This PR also adds NPM caching to all the matrix environments.

- [x] Blocked until #212 is merged.

## How?

This PR does the following:

- Extracts the Node.js and PHP environment configurations into custom actions, `.github/actions/node` and `.github/actions/php` respectively. Reduces duplicate code when setting up multiple workflows needing the same steps.
- Set up caching of Node.js dependencies, which should speed up CI workflows as long as the `package-lock.json` doesn't change. It works exactly the same to the previous caching of `composer` dependencies. 
- Uses the new environment setup actions to set up three new workflows:
  - `node-build.yml` - Check that the Node.js portion of the project builds without error.
  - `test-suite.yml` - Check that the Node.js and PHP tests pass.
  -  `coding-standards.yml` - Check coding standards of JavaScript and PHP files.

PR #219, opened 2023-04-03
chore: update `@wordpress/scripts`

johnhooks:
## What?

Update `@wordpress/scripts` to latest, `v26.1.0`.

## Why?

Trying to diagnosis Issue in the ESLint configuration discovered in PR #213.


erikyo (approved):
LGTM 🎉

erikyo (dismissed):

johnhooks:
@erikyo this doesn't fix the issue yet. We'll merge it in once we actually discover the problem.

johnhooks:
I do think all the additions of this PR at the moment are a good idea, but they don't fix the Prettier ESLint reporting.

erikyo:
> @erikyo this doesn't fix the issue yet. We'll merge it in once we actually discover the problem.

@johnhooks this is fixing also some issues that i got merging the storybook PR, since are package related

johnhooks:
@erikyo do we want to merge it in? And I'll keep looking and make another PR when I find it?

I moved a lot of information about this to Issue #220.

If you want to merge, go ahead and we will keep looking.

erikyo:
@johnhooks I'll try to do cherry picking from this in the meantime and if I can if you think it's not ready to be merged yet, but ideally it's okay for me since it's updating old packages. 

Only one thing... there would be no need to add packages like "@wordpress/eslint-plugin" because they are already bundled with wp-scripts, check this 
https://github.com/WordPress/performance/pull/468#discussion_r1024658920

johnhooks:
@erikyo It was mostly a Hail Mary attempt, I can remove

erikyo:
Yes, if it is possible, we should only add packages that are not already available, so we avoid having to maintain them and we are sure they are the "right ones"

https://github.com/WordPress/gutenberg/blob/641949ab72f65e617d3bf69cb71568cfe6799188/packages/scripts/package.json#L60-76

erikyo (approved):

erikyo (approved):

PR #221, opened 2023-04-03
fix: previous files interpolation in .lintstagedrc

johnhooks:
## What?

fix: previous files variable interpolation in .lintstagedrc

## Why?

The lint-staged config for prettier is broken.


erikyo (approved):
LGTM 🎉

Issue #220, opened 2023-04-03
ESLint configuration issue


johnhooks:
### Description

Fix issue in ESLint configuration.

### Why?

Issue in the ESLint configuration discovered in PR #213 .

### Solutions attempted

The following did not resolve the issue in PR #219 

- [x] Installing `@wordpress/eslint-config` and `@wordpress/prettier-config` peer dependencies.
- [x] Blowing away the `package-lock.json` and reinstalling.

### Step-by-step reproduction instructions

`npm run lint:js`

### Please confirm that you have searched existing issues in the repo.

Yes

### Please confirm that you have tested with all plugins deactivated except WP Feature Notifications.

Yes

PR #233, opened 2023-04-04
fix: eslint issue

johnhooks:
## What?

Fix ESLint / Prettier configuration.

Fixes Issue #220 

## Why?

Apparently there is something called `wp-prettier` which is a maintained fork of prettier, just to add WP styling. There is a whole thing, if you want to dive deep here are some links:

- [Prettier Issue #13107](https://github.com/prettier/prettier/issues/13107)
- [wp-prettier](https://www.npmjs.com/package/wp-prettier)

I think some other dependency must have overridden the overriding of `prettier` by `@wordpress/scripts`'s `wp-prettier`. But now the package.json explicitly states we are using `wp-prettier` in place of `prettier`.


erikyo (approved):
LGTM 🎉

PR #236, opened 2023-04-04
update to react 18 (converts createElement to createRoot)

### LINKED ISSUE #164: Upgrade to React 18
### Description

ReactDOM.render has to be replaced because it is no longer supported in React 18. 

https://reactjs.org/blog/2022/03/08/react-18-upgrade-guide.html#updates-to-client-rendering-apis

### Step-by-step reproduction instructions

Run the command `npm run packages-update` and (assuming you are using wp-env) navigate to "http://localhost:8889/wp-admin", log in and check the console. 

An error message will be displayed in the console like this: "Warning: ReactDOM.render is no longer supported in React 18. Use createRoot instead. Until you switch to the new API, your app will behave as if it's running React 17. Learn more: https://reactjs.org/link/switch-to-createroot" 

### Screenshots, screen recording, code snippet

_No response_

### Environment info

WP-env -> {
	"core": "WordPress/WordPress",
	"plugins": ["."]
}

Yes

### Please confirm that you have tested with all plugins deactivated except WP Feature Notifications.

Yes

linked to: #236

erikyo:
## What?
An error message will be displayed in the console like this: "Warning: ReactDOM.render is no longer supported in React 18. Use createRoot instead. Until you switch to the new API, your app will behave as if it's running React 17. Learn more: https://reactjs.org/link/switch-to-createroot"
ReactDOM.render has to be replaced because it is no longer supported in React 18.

## Why?
Run the command npm run packages-update and (assuming you are using wp-env) navigate to "http://localhost:8889/wp-admin", log in and check the console.

closes #164


Sephsekla:
This is good, but I'm also conscious it makes 6.2 our minimum supported WP version, which might limit people testing it at this stage. 🤔

erikyo:
> This is good, but I'm also conscious it makes 6.2 our minimum supported WP version, which might limit people testing it at this stage. 🤔

good catch but since we are already using wp-scripts 26.x.x, we are also yet using React 18. Maybe we should decide what policy to adopt here, but IMO we need to stay aligned with the latest version

johnhooks:
@Sephsekla if we get a WordPress playground working that makes it easy to demo the plugin, would you be more comfortable with this approach?

Sephsekla:
OK, I think given that we're using the newest version of wp-scripts it does make sense to be forward-thinking here - but it would be good to discuss version support today and tomorrow and I can ensure that it's addresssed in the post that will go out as soon as we have the release ready.

Sephsekla:
Could you also do the `packages-update` as part of this PR? Makes most sense to me.

Sephsekla:
Likewise, as discussed if you can update the minimum supported WP version in the plugin to go with this, then that should keep things clear.

.wp-env.json
```diff
@@ -1,4 +1,4 @@
 {
-	"core": "WordPress/WordPress",
-	"plugins": ["."]
+	"core": null,
```
Sephsekla: I think we can most likely remove this altogether, given it's the default. Not a blocker though.

Sephsekla (dismissed):
🚀

.wp-env.json
erikyo: yes actually it doesn't make sense, if you want we can remove it in this PR since we are updating it, okay?

erikyo:
```suggestion
```

Sephsekla (approved):

PR #237, opened 2023-04-04
removes unnecessary packages

erikyo:
## What?
remove some packages that aren't needed anymore (assuming that in the past they have been)

## Why?
I am upgrading to Storybook 7 and I realized that a future we will have problems with these because of un-synced packages required by wp-script. And furthermore, since they are already available in the wp-script package.json why include them?

## How?
Removed packages from packages.json

## Testing Instructions
wipe the node_packages folder then `npm i`.  In order to test give a try to start/build the plugin scripts


johnhooks (dismissed):
LGTM

erikyo:
I wanted to give it a try with the new CI workflow, and I'd say it's working fine! thanks @johnhooks 🤣

johnhooks (approved):

Sephsekla (approved):

PR #240, opened 2023-04-05
187 Release action

### LINKED ISSUE #187: Create a GitHub action to build scripts and styles when a new release is tagged

Currently, scripts and styles are manually built, and committed to the repo. This is not ideal.

@JasonTheAdams has kindly provided a good starting point at https://github.com/impress-org/givewp-github-actions/blob/59fcb8b536bd4955c6fa7319fac7f0c0558408d4/.github/workflows/givewp-release.yml#L129-L137. This will need updating for our workflow.

Sephsekla:
## What?

Create a GitHub action to create a release for a tag, and upload asset files.

## Why?

Resolves #187 

## How?

This PR creates a new GitHub workflow, which will run when a tag is created:

- Checkout the tag
- Setup node (as in the build workflow)
- Build the assets
- Create a zip folder (excluding certain files using .distignore)
- Create a new release for the tag and attach the zip


johnhooks:
~@Sephsekla what are the conditions you would like to trigger this workflow?~ Scratch that, I took a look at the workflow file.

Sephsekla:
@johnhooks right now I'm thinking that on tag, we should checkout the code, build, remove certain folders (e.g. node_modules, .github), and then zip.

Then create a new release and attach using https://github.com/softprops/action-gh-release

johnhooks:
Sounds like a solid plan to me. I'm working on a WordPress Playground, it could be redeployed in this same workflow.

Aside: It will be nice to have `build` removed.

johnhooks (approved):
@Sephsekla This looks great.

Is the PHP environment setup required to properly build the release? As far as I can tell it isn't, but wanted to double check.

Sephsekla:
> Is the PHP environment setup required to properly build the release? As far as I can tell it isn't, but wanted to double check.

At the moment we only use composer for devDependencies and testing, so it shouldn't be necessary for release. That might change in future though, we'll need to keep an eye out.

PR #244, opened 2023-04-05
Update dependabot config to reduce noise

### LINKED ISSUE #149: set Dependabot only for security updates
What we are interested in is that there aren't any serious security problems, if we get so many pr for each updated package we risk losing sight of the real pr

According [to this article](https://docs.github.com/en/code-security/dependabot/dependabot-security-updates/configuring-dependabot-security-updates) just set `open-pull-requests-limit` to 0 [here](https://github.com/WordPress/wp-feature-notifications/blob/bac7a329597e2ddc8257d8c51c5f5f2fcc5b686a/.github/dependabot.yml#L9) for disabling version updates for npm dependencies

linked to #244

Sephsekla:
## What?

Make several updates to Dependabot config to reduce the frequency of notifications.

## Why?

Resolve #149

## How?

Make several updates to Dependabot config:

- Dependabot PRs will now be created weekly, rather than daily.
- Dependabot will only create updates for direct dependencies, and also any `@wordpress` dependencies, however this may update other dependencies transitively.
- Security PRs will still be created as before.

This should hopefully make this more manageable without sacrificing security or maintenance.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,24 +1,28 @@
 version: 2
 
 updates:
-  - package-ecosystem: "composer"
-    directory: "/"
-    schedule:
-      interval: "daily"
-    versioning-strategy: lockfile-only
-    open-pull-requests-limit: 10
-    commit-message:
-      prefix: "Composer"
-    allow:
-      - dependency-type: "all"
+    - package-ecosystem: 'composer'
+      directory: '/'
+      schedule:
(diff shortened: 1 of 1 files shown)


johnhooks (approved):
This looks like it should reduce the quantity of bot PRs. Let's give it a try.

erikyo (approved):

PR #245, opened 2023-04-06
refactor: convert to namespaces

### LINKED ISSUE #197: Create naming convention

johnhooks:
## What?

Implement naming convention to discussed in Issue #197.

## Why?

There is currently a few differently systems of naming across the project and it should be consistant.

## How?

- Remove `WP_Notify` class prefix
- Remove `wp-notify` file prefix
- Convert to namespace `WP\Notifications`

Also cleans up some misaligned PHPDoc comments.


wp-feature-notifications.php
```diff
@@ -15,52 +15,53 @@
...
-	define( 'WP_NOTIFICATION_CENTER_PLUGIN_DIR', dirname( __FILE__ ) );
+if ( ! defined( 'WP_FEATURE_NOTIFICATION_PLUGIN_DIR' ) ) {
+	define( 'WP_FEATURE_NOTIFICATION_PLUGIN_DIR', dirname( __FILE__ ) );
 }
 
-if ( ! defined( 'WP_NOTIFICATION_CENTER_PLUGIN_DIR_URL' ) ) {
-	define( 'WP_NOTIFICATION_CENTER_PLUGIN_DIR_URL', plugin_dir_url( __FILE__ ) );
+if ( ! defined( 'WP_FEATURE_NOTIFICATION_PLUGIN_DIR_URL' ) ) {
+	define( 'WP_FEATURE_NOTIFICATION_PLUGIN_DIR_URL', plugin_dir_url( __FILE__ ) );
 }
```
johnhooks, line 32: @Sephsekla the globals that call functions can't be namespaced `const` variables.

johnhooks (commented):
This is a first step to fix naming in the PHP code. The only code removed are the serialization tests. They are problematic. The previous system of serializing class instances into JSON that is initialized based on class names embedded into the JSON won't fit well with standard WordPress rest-api schemas.

includes/class-factory.php
```diff
@@ -1,35 +1,42 @@
 <?php
 
-class WP_Notify_Factory {
+namespace WP\Notifications;
+
+use WP\Notifications\Messages\Message_Factory;
```
Sephsekla: It's a style thing really, but I personally always prefer to import namespaces rather than specific classes  - keeps things clearer, and also lets you name things a bit more concisely.

So rather than having 

```php
use WP\Notifications\Recipients\Recipient_Factory;
use WP\Notifications\Recipients\Recipient_Collection;

//...

/**
 * Recipient factory implementation to use.
 *
 * @var Recipient_Factory
 */
private $recipient_factory;

//...

$recipients = new Recipient_Collection();
```

You could rename the class and simplify imports to something like

```php

use WP\Notifications\Recipients;

//...

/**
 * Recipient factory implementation to use.
 *
 * @var Recipients\Factory
 */
private $recipient_factory;

//...

$recipients = new Recipients\Collection();
```

wp-feature-notifications.php
Sephsekla: Ah of course 🤦

Sephsekla (commented):
In general there's a lot that needs doing still in terms of documentation etc - there are a lot of classes etc which need commenting (and as discussed, maybe some things we can remove).

Added one thought around namespace imports, let me know what you think.

includes/class-factory.php
johnhooks: I am really new to PHP namespaces, so I don't have any preferences. This works for me!

I'll make this change.

One question, should classes in the global namespace be imported at the top of a file with:

```
use Exception;
```

or done inline?

```
\Exceptions
```

johnhooks:
@Sephsekla I agree, this is just a first step. How would you like to work through the whole process? 

I would prefer doing it in stages so each PR is simple in purpose and easier to review. This one is rather large, but I think its narrow scope makes it easy to review and understand.

wp-feature-notifications.php
johnhooks: I left the `WP_FEATURE_NOTIFICATION_PLUGIN_VERSION` as a defined global rather than const for consistency, do you agree?

includes/class-factory.php
johnhooks: I went with the use statement option for globals

johnhooks (commented):
I checked everything and this is ready for review.

includes/demo.php
```diff
@@ -36,27 +41,27 @@ function wp_admin_bar_wp_notify_item( WP_Admin_Bar $wp_admin_bar ) {
...
 	echo '<div id="wp-notify-dashboard" class="wrap"></div>';
 }
-add_action( 'admin_notices', 'wp_notify_admin_notice' );
+add_action( 'admin_notices', '\WP\Notifications\admin_notice' );
 
 /**
  * Register and enqueue a wp notify scripts and stylesheet in WordPress admin.
  */
-function wp_notify_enqueue_admin_assets() {
+function enqueue_admin_assets() {
```
erikyo, line 57: the changes in this files are already fixed in #252 , thanks for not overlapping your changes

johnhooks: Do you want me to remove the changes to this file? Or will it in its current version not effect #252

erikyo (approved):
I have searched around all files but there is no more trace of wp-notify! Good job @johnhooks and thanks, now we have all the files consolidated 🚀

PR #246, opened 2023-04-07
fix(js): notices loop item key

johnhooks:
## What?

Use the notice's id as the component key rather than the array index.

## Why?

React uses the key to determine how to update the DOM, if the index is used as a key, and an item is removed from the middle of the array. React doesn't know about it and removes the last element in the list and leaves the element that was intended for removal in the DOM.


erikyo (approved):
good catch! thank you @johnhooks

PR #248, opened 2023-04-07
build: add typescript

johnhooks:
## What?

Add `typescript` to enable JSDoc typing of the JavaScript code base. The `@wordpress/eslint-plugin` docs state that adding `typescript` will enable the proper ESLint configuration for importing JSDoc types between files.

## Why?

Proper typing greatly enhances IDE autocomplete and API discovery .

Typing will also help catch errors.


erikyo (approved):
LGTM 🚀

PR #250, opened 2023-04-09
feature: add script to generate/refresh demo data

johnhooks:
## What?

The datetimes of the demo data were old, and I wanted to have a way to refresh them. This script starts the work on creating a set of scripts to generate data for the demo.

## How?

Add a scripts directory and calls the `gen-demo-data` script on `prepare` so the data is generated when running `npm install`. Also moves the generated test data to a `tmp` folder so it is outside of source control.


johnhooks:
This will be performed by `wp-cli` custom commands which are included in #259.

PR #251, opened 2023-04-10
feature: add channel registry

johnhooks:
## What?

Add a `Channel_Registry` class to register notification channels.

## Why?

To attempt registering channels in code.

Implementing the data model of PR #168 has the challenge of when and how to look up plugin registered channels. If channel data is stored in a database table and an integer `id` column is used to link a channel to subscription and messages, how does a plugin know the `id` of the channel it registered? This data would have to be looked up on message emit, and also there wouldn't be a good mechanism to prevent plugins from registering similarly named channels.

Using a registry similar block types might work well. The channel is registered in code and the channel name/key/slug (as in `core/post-edit`) would link a message source to its channel. Subscriptions could also use the same key to link users to channels. This would also allow checking for conflicts at channel registration.

Keeping in mind that if/when a plugin is uninstalled, a message has to contain all the data required to render, it is necessary to store channel key and title in the message.

Both systems have pros and cons. This strategy removes the need for a channel table, and simplifies the logic of emitting a message. Though it will result in duplication of data in the database.

## How?

- `Channel_Registry` class, heavy based on core `Block_Type_Registry`
- `Channel` class
- `register_channel` function

## Further considerations

- When/if a plugin is uninstalled its channel subscriptions would remain in the database subscriptions table. Though this is a very small table compared with the queue table. And remembering user preferences is probably normal behavior, so there is a seamless experience if a plugin is reinstalled.
- The channel `icon` could be stored in the message's `icon` column if/when a message doesn't specify it's own.

### DIFF
--- a/includes/channels.php
+++ b/includes/channels.php
@@ -8,14 +8,18 @@
 namespace WP\Notifications;
 
 /**
- * Register a channel.
+ * Register a notification channel.
  *
- * @param Channel $channel A complete Channel instance.
- *
+ * @param string|Channel $name  Channel name including namespace, or alternatively a complete
+ *                              Channel instance. In case a Channel is provided, the $args
+ *                              parameter will be ignored.
+ * @param array          $args  Optional. Array of channel arguments. Accepts any public property
+ *                              of `Channel`. See Channel::__construct() for information on
+ *                              accepted arguments. Default empty array.
  * @return Channel|false The registered channel on success, or false on failure.
  */
-function register_channel( $channel ) {
-	return Channel_Registry::get_instance()->register( $channel );
+function register_channel( $name, $args = array() ) {
+	return Channel_Registry::get_instance()->register( $name, $args );
 }
 
 /**
--- a/includes/channels.php
+++ b/includes/channels.php
@@ -21,12 +21,10 @@ function register_channel( $channel ) {
 /**
  * Unregister a channel.
  *
- * @since 5.0.0
- *
  * @param string|Channel $name Channel name including namespace, or alternatively a complete
  *                             Channel instance.
  * @return Channel|false The unregistered channel on success, or false on failure.
  */
-function unregister_block_type( $name ) {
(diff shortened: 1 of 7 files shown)


includes/channels.php
```diff
@@ -0,0 +1,32 @@
...
+/**
+ * Unregister a channel.
+ *
+ * @since 5.0.0
+ *
+ * @param string|Channel $name Channel name including namespace, or alternatively a complete
+ *                             Channel instance.
+ * @return Channel|false The unregistered channel on success, or false on failure.
+ */
+function unregister_block_type( $name ) {
```
johnhooks: I need to fix this, should be `unregister_channel`

johnhooks:
The `register_channel` function seems awkward. It could look cleaner to pass an array with channel constructor parameters.

Though this achieves creating a channel with translations for title and description.

https://github.com/WordPress/wp-feature-notifications/blob/a33543cc448ac6ed29a5cb667fb79682a9fbf019/wp-feature-notifications.php#L74-L81

johnhooks:
This looks a lot cleaner to me, though PHP doesn't help with type hinting the array. Named arguments would be most preferred, but that would require a minimum PHP version of 8.

https://github.com/WordPress/wp-feature-notifications/blob/cb4954ea3f5ecf140ac96194201354deb4ea76cf/wp-feature-notifications.php#L77-L86

includes/class-channel.php
```diff
@@ -0,0 +1,85 @@
...
+	 * @var string|null
+	 */
+	public $description = null;
+
+	/**
+	 * Constructor.
+	 *
+	 * Instantiates a Channel object.
+	 *
+	 * @see WP\Notifications\register_channel()
```
erikyo: wasn't something like this?
`@see \WP\Notifications\register_channel()`

johnhooks:
Blocked by #255

johnhooks:

PR #252, opened 2023-04-10
Drawer update

### LINKED ISSUE #182: The shadow in side bar when no overflow
### Description

When there isn't overflow in the side panel there shouldn't be the shadow, as described in the design documents.

### Step-by-step reproduction instructions

1. Go to the notifications side panel
2. Dismiss all notifications
3. The shadow is still present near the bottom of the panel.

### Screenshots, screen recording, code snippet

Screen Shot 2023-03-24 at 7 08 09 PM

### Environment info

- Run with the latest develop branch and wp-env

Yes

### Please confirm that you have tested with all plugins deactivated except WP Feature Notifications.

Yes

### LINKED ISSUE #186: Add the ability to expand the Notification Hub

The Notification Hub is rendered at a fixed width and some of the notification messages feel tight, which could even worse in other languages.

Adding the ability to drag the sidebar edge and widen the Hub area would provide the user more control over how they view their notifications.

erikyo:
## What?
This is a complete refresh of the drawer which is now a React component. 

## Why?
Previously the drawer was created by adding some html markup next to the menu during the initialization of the admin bar. This allowed us to have already in place an area where to initialize the component to display notifications (with the context of the admin-bar) but had also some drawbacks e.g. the sidebar that could not be a component.

## How?
We revised and rewrote some js scripts, plugin init, added/updated components, adjusted css styling.
@johnhooks provided types and fixed some reducer related issues

## Screenshots or screencast
resize

closes #186, fix #182


src/scripts/components/Notice.js
```diff
@@ -9,38 +9,38 @@ import { NoticeIcon } from './NoticeImage';
...
 export const Notice = ( props ) => {
 	const {
-		id,
-		title,
-		status,
+		action,
 		context = defaultContext,
-		source = 'WordPress',
-		date = __( 'Just now' ),
+		date = Date.now() * 0.001,
```
johnhooks, line 34:
```suggestion
		date = Math.floor(Date.now() / 0.001),
```
The data should be an integer.

johnhooks:
```suggestion
		date = Math.floor(Date.now() * 0.001),
```

The date should be an integer.

src/scripts/components/NoticesLoop.js
```diff
@@ -4,10 +4,15 @@ import { Notice } from './Notice';
  * Returns a list of notices
  * each notice is a component that has a key, an id, an image, an additional class name, and an onDismiss function
  *
- * @param {Object} prop
- * @param {Array}  prop.notices - An array of objects that contain the notice data.
+ * @param {Object}                      prop
+ * @param {import('../store').Notice[]} prop.notices - An array of objects that contain the notice data.
```
johnhooks, line 8: It would be preferable to import this at the top of the file. It provides a cleaner looking function signature

```suggestion
 * @param {Notice[]} prop.notices - An array of objects that contain the notice data.
```

like:

```js
/**
 * @typedef {import('../store').Notice} Notice
 */
```
```

**Review by johnhooks — CHANGES_REQUESTED** [review]

File: src/scripts/components/NoticesLoop.js (hunk 2 above)
Comment (johnhooks) [reply src/scripts/components/NoticesLoop.js:8]: I realize I'm the one that wrote the original, and now I remember why it is like that. There is a name conflict between `Notice` as a type and as a react component if the type is imported like a suggested. Leaving it as is would be best.

File: src/scripts/components/Notice.js [hunk 3]
```diff
@@ -9,38 +9,38 @@ import { NoticeIcon } from './NoticeImage';
...
 export const Notice = ( props ) => {
 	const {
-		id,
-		title,
-		status,
+		action,
 		context = defaultContext,
-		source = 'WordPress',
-		date = __( 'Just now' ),
+		date = Math.floor(Date.now() * 0.001),
```
erikyo:
```suggestion
		date = Math.floor( Date.now() * 0.001 ),
```

erikyo: Just noticed @johnhooks have fixed in a better way 🚀. resolved!

johnhooks (dismissed):
LGTM

src/scripts/components/NoticesLoop.js
bacoords: Since there is a [`Notice` component](https://wordpress.github.io/gutenberg/?path=/docs/components-notice--default) in Gutenberg already- is the plan to rename this or to eventually use the core `Notice`?

johnhooks: Probably not for use in this component, though this could be renamed to be more specific. I believe the Gutenberg `Notice` is a toast notification and is much simpler. @erikyo and I have talked about supporting toast notifications, and I see a real benefit in using the components of Gutenberg. Though we should use the same design system if we intend to use Gutenberg components along side our own.

erikyo: Agree! Also we would have to heavily modify the notifice component to make it look like how we need it and I don't know if it's worth it... if I'm not mistaken [this one](https://wordpress.github.io/gutenberg/?path=/docs/components-notice--default) does not foresee the use of an image or a title for example, and even aesthetically similar we can hardly use it!

bacoords: Yeah I agree that the core component would have to be heavily updated to work with this project- images, action items, etc- but at some point I could imagine that we wouldn't want two components with the same name.

Core does have a "[Snackbar](https://wordpress.github.io/gutenberg/?path=/story/components-snackbar--default)" which is essentially the toast component you mentioned. I agree, that could be a benefit to incorporate at some point.

erikyo: yes indeed! with @johnhooks we identified also another additional "notification type", the **modal window** that isn't covered by the original mockup of wp-notification too (but somehow present in Gutenberg components).

johnhooks: @bacoords ah you are right, I was thinking of the Snackbar. And I agree, we should work to expand `@wordpress/components` and contribute back to reduce duplication and keep building out the component library. Though I think we have a lot of work to do and need to figure out what works first.

johnhooks (dismissed):

includes/demo.php
```diff
@@ -35,7 +35,7 @@ function admin_bar_item( WP_Admin_Bar $wp_admin_bar ) {
 /**
  * Adds WP Notify area at the top of the dashboard
```
johnhooks: Can you remove "notify" from all the comments?

johnhooks (changes requested):

includes/demo.php
erikyo: it still remains "wp-notify" in some css for custom props (and maybe few other css comments) but we will make a pr specifically for that! closing this

johnhooks (approved):
LGTM

PR #255, opened 2023-04-12
ci: remove php 7.2 add 8.1 and 8.2

johnhooks:
## What?

Remove PHP 7.2 from CI `test-suite.yml` workflow.

Add PHP 8.1 and 8.2 to CI `test-suite.yml` workflow.

Also `lint-staged` formatted the yaml files to match the prettier rules for yaml, which uses 4 spaces for tabs.

Also remove `set-output`, there was a warning it is depreciated and will be removed soon.

## Why?

The plugin header specifies minimum PHP support is 7.4.

### DIFF
--- a/.github/actions/php/action.yml
+++ b/.github/actions/php/action.yml
@@ -1,29 +1,29 @@
 name: Setup PHP
 
 runs:
-  using: 'composite'
-  steps:
-    - name: Setup PHP
-      uses: shivammathur/setup-php@v2
-      with:
-        php-version: ${{ matrix.php }}
-        tools: composer
-        extensions: mysql
-        coverage: none
+    using: 'composite'
+    steps:
+        - name: Setup PHP
+          uses: shivammathur/setup-php@v2
+          with:
+              php-version: ${{ matrix.php }}
+              tools: composer
+              extensions: mysql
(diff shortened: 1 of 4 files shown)


erikyo (dismissed):
LGTM 🚀

johnhooks:
@Sephsekla we need your review on this one.

Sephsekla (dismissed):
I am slightly wary that we'll need to enable support for older versions somewhere down the line due to WordPress' own support for *ancient* stuff. But for now happy to drop 7.2.

Can we add 8.2? Should be supported in beta as of WordPess 6.1.

johnhooks:
@Sephsekla absolutely I will add 8.2. Also I think you can feel comfortable dropping 7.2, WP's minimum requirement right now is 7.4
https://wordpress.org/about/requirements/

Sephsekla (dismissed):

johnhooks:
@Sephsekla I can't squash and merge it. Do you think @erikyo and I could have permissions to override the checks? They are super helpful in most context, but there have been a lot of situations like this that we haven't be able to override them when necessary.


```
Required statuses must pass before merging
```

Sephsekla (approved):

PR #256, opened 2023-04-12
build: update composer packages

johnhooks:
## What?

Update composer.json and lock file.

## Why?

The packages need to be updated.

~Blocked by PR #255~


erikyo (approved):
LGTM 🚀

PR #269, opened 2023-04-14
build: make storybook a workspace

johnhooks:
## What?

Simply the dependency management of storybook by adding to the project as a workspace.

## Why?

- Running install in the root directory will install dependencies for storybook.
- Allows running scripts from the workspace.


package.json
```diff
@@ -23,9 +23,14 @@
...
 	],
+	"workspaces": [
+		"./storybook"
+	],
 	"scripts": {
 		"start": "wp-scripts start",
+		"start:storybook": "npm run start --workspace=storybook",
 		"build": "wp-scripts build",
+		"build:storybook": "npm run build --workspace=storybook",
 		"prephp-install": "wp-env start",
```
erikyo: i've noticed also there two that are "duplicated", is worth to remove them in this PR? If yes do it, however check that they are not used in the php test workflow

johnhooks: @erikyo which two are duplicated?

erikyo (approved):
LGTM 🎉

PR #270, opened 2023-04-14
ci: deploy storybook docs on release

johnhooks:
## What?

Add deployment of Storybook documentation to GitHub Pages in release workflow

## Why?

Keep documentation current with the latest release.

**Additionally**

- Fix `PLUGIN_SLUG` shell variable in release workflow.
- Exclude all unnecessary project files from the release zip in the `.distignore`.
- Add input to actions/node and not use the matrix variable directly inside the action.
- Reformat yaml files to comply with prettier rules.


johnhooks:
@Sephsekla I tested this all out in my own repo

- The workflow runs successfully.

   https://github.com/johnhooks/wp-feature-notifications/actions/runs/4705531604

- The plugin zip is added to the release assets.

   https://github.com/johnhooks/wp-feature-notifications/releases

  The title and release message need work, we should add a changelog. If we adopt conventional commits we can auto generate the changlog for a release based on the commit message.
  
   The `wp-feature-notifications.zip` in the release didn't have the `src` directory excluded, I've added it to the`.distignore` and won't be included in release plugin zips.

- The docs are build at deploy to the gh-pages branch.

   https://github.com/johnhooks/wp-feature-notifications/tree/gh-pages

- With the right settings the docs are hosted through GItHub Pages.

   https://johnhooks.io/wp-feature-notifications/

There are still some issues with the deployed Storybook site, but it is really close to automated deployment. GitHub Pages will need to be enabled on the `gh-pages` branch once this workflow actually runs in the repo.

.distignore
```diff
@@ -1,2 +1,49 @@
...
+
+babel.config.js
+composer.json
+composer.lock
+jest.config.js
+package-lock.json
+package.json
+phpcs.xml.dist
+phpunit.xml.dist
+README.md
```
Sephsekla, line 44: I think it's worth having a readme in the distribution zip, personally, although I agree with the rest. What do you think?

Sephsekla (dismissed):
Nice work! One thought around what's included in the release zip, otherwise looks great.

.distignore
erikyo: What is excluded is not the readme.txt (the one used by WordPress) but the one that is used by GitHub, ad IMHO it is ok like this because otherwise it would be a "duplicate"

Sephsekla: 🤦 oh that's a good point. Feel free to disregard, I think that makes sense.

johnhooks: @Sephsekla I drew a lot from the GiveWP repo, they exclude the readme markdown file too.

erikyo (approved):
LGTM! 🎉

PR #271, opened 2023-04-14
Provides a default value to the drawer

erikyo:
## What?
the Drawer component requires some mandatory props

## Why?
We can set a default value to be able to call the components without adding any arguments, so it will be easier to import

## How?
adding the default values like 
```js
const NotificationHub = ( {
	active = false,
	initialActive = false,
} ) => {'...'}
```


johnhooks (approved):
LGTM. If we want to use the other hook we need to find a better way to import it.

PR #273, opened 2023-04-15
CSS - Moving to "purpose naming" convention

erikyo:
## What?
Css custom props were definitely a great innovation and allow you to share the colors of your stylesheet. However, too many plugins end up creating their own color set libraries in the root and you end up with the opposite effect

## Why?
IMHO the problem arise because the name of the color is used to set the variable name, and this eventually forces all plugins to create a variable for each color used. This PR makes this plugin follow another method which is to use the names on the purpose that the variables have (and thus success, error etc)
Several articles that talk about this can be found (eg. [this](https://engineering.monday.com/how-we-stopped-using-color-names-and-move-to-purpose-naming/) or [this](https://www.freecodecamp.org/news/css-naming-conventions-that-will-save-you-hours-of-debugging-35cea737d849/)), it seems to me a very good improvement to do to clean up and simplify the style

## How?
```scss
:root {
  --wp-notify--color--red: #f00; // Red
  --wp-notify--color--error: var(--wp-notify--color--red);
}
```
to
```scss
$color-red: #d94f4f;
:root {
  --wp-notify--color--error: #{$color-red};
}
```

### DIFF
--- a/src/styles/components/bullet.scss
+++ b/src/styles/components/bullet.scss
@@ -4,7 +4,6 @@
 
 $bullet-width: 6px;
 
-
 #wpbody .wp-notification-wrap,
 #wpadminbar .wp-notifications.new {
 	position: relative;
--- a/src/styles/components/notification.scss
+++ b/src/styles/components/notification.scss
@@ -2,8 +2,8 @@
  * General notifications style
  */
 .wp-notification {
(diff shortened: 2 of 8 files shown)


johnhooks (approved):
This looks good. Eventually it would be nice if we could import the same colors used by Gutenberg.

PR #274, opened 2023-04-15
Move to @wordpress/escape-html and @wordpress/date packages

erikyo:
## What?
Initially (out of habit) I had used the Moment and domPurify packages, but similar things already exist in the Gutenberg repo and it is definitely better to use these, since now the compiled sources weigh a tenth as much as before.

## How?
I make amends and this PR updates to WP built in packages

### DIFF
--- a/src/scripts/components/NoticeMeta.js
+++ b/src/scripts/components/NoticeMeta.js
@@ -1,4 +1,4 @@
-import { dateI18n } from '@wordpress/date';
+import { noticeDateFormat } from '../utils';
 
 /**
  * The notice metadata, for example the source of the notification or the date
@@ -10,8 +10,6 @@ import { dateI18n } from '@wordpress/date';
 export const NoticeMeta = ( { date, source } ) => (
 	<p className="wp-notification-meta">
 		<span className="name">{ source }</span> { '\u2022 ' }
-		<span className="date">
-			{ dateI18n( 'l jS F Y - h:i A', date, true ) }
-		</span>
+		<span className="date">{ noticeDateFormat( date ) }</span>
 	</p>
 );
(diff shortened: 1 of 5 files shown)


src/scripts/components/NoticeMeta.js
```diff
@@ -1,15 +1,17 @@
...
- * @param {number} props.date   The date of the notification.
+ * @param {string} props.date   The date of the notification.
  * @param {string} props.source The source of the notification.
  */
 export const NoticeMeta = ( { date, source } ) => (
 	<p className="wp-notification-meta">
 		<span className="name">{ source }</span> { '\u2022 ' }
-		<span className="date">{ moment( date ).fromNow() }</span>
+		<span className="date">
+			{ dateI18n( 'l jS F Y - h:i A', date, true ) }
```
johnhooks: Can you make a utility function that does this named `fromNow` or `timeFromNow`?

erikyo: Actually just converts the epoch to a readable time... Would it be better `toCurrentDate`?

src/scripts/components/NoticeMeta.js
```diff
@@ -1,15 +1,15 @@
-import * as moment from 'moment/moment';
+import { noticeDateFormat } from '../utils';
 
 /**
  * The notice metadata, for example the source of the notification or the date
  *
  * @param {Object} props
- * @param {number} props.date   The date of the notification.
+ * @param {string} props.date   The date of the notification.
```
johnhooks: The date is a string?

erikyo: seems the [dateI18n accepts](https://developer.wordpress.org/block-editor/reference-guides/packages/packages-date/#datei18n) only stings and not numbers

johnhooks (approved):
I think these are great changes and really reduce the size of this package and uses the WordPress provided functionality.

PR #275, opened 2023-04-15
feature: use datetime strings in fake api data

johnhooks:
## What?

Use the standard datetime strings used in the other REST API endpoints of WordPress in the fake api data.

Convert datetime strings into JS `Date`s at the time they are fetched from the API.

## Why?

Using the standard datetimes makes the fake api more like the real thing.

**Additions**

- Move the some of the constants out of the store folder, they are used in many place in the project.
- Use `/** @type {const} */` type assertions, JSDoc style `as const`.


erikyo (approved):
LGTM!

PR #277, opened 2023-04-16
refactor: Drawer width logic

johnhooks:
## What?

Prevent the Drawer component from becoming larger than the window area.

## Why?

Previously the Drawer could become larger than the window area by a user dragging the edge outside the window, or the window being resized smaller than the Drawer. Then the edge of the drawer was no longer in view and the drawer was stuck in a size too large to fit in the window.

## How?

Added logic:

- To prevent the Drawer from being resized larger than the window when dragging the handle.
- To resize the Drawer on window resize if its width is smaller than the window. Though it will not go lower than minimum `HUB_WIDTH`.

### DIFF
--- a/src/scripts/components/Drawer.js
+++ b/src/scripts/components/Drawer.js
@@ -1,4 +1,4 @@
-import { useState } from '@wordpress/element';
+import { useState, useEffect, useCallback } from '@wordpress/element';
 import { __ } from '@wordpress/i18n';
 import { useShortcut } from '@wordpress/keyboard-shortcuts';
 import { Resizable } from 're-resizable';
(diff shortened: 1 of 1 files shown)


src/scripts/components/Drawer.js
```diff
@@ -33,8 +58,13 @@ export const Drawer = ( { focus, blur, instance } ) => {
...
-					setWidth( width + d.width );
+				onResizeStop={ ( _e, _direction, _ref, d ) => {
+					const currentWidth = width + d.width;
+					if ( currentWidth > window.innerWidth ) {
+						setWidth( window.innerWidth );
+					} else {
+						setWidth( width + d.width );
+					}
 				} }
 				minWidth={ HUB_WIDTH }
```
erikyo, line 43: since the drawer during resize could be larger than the viewport could you kindly add this (seems to fix the issue)

resize

```js
maxWidth={ '100vw' }
```

erikyo: furthermore ( can be seen in the gif above) we need to set something (like below) because the padding seems to be added to the width of the drawer

```css
#wp-notifications-hub {
  box-sizing: border-box;
}
```

erikyo (dismissed):
LGTM!

src/scripts/components/Drawer.js
erikyo: scratch this ☝️about the box-sizing... maybe it needs a drawer css style rework and is better to do it in a different pr

johnhooks: It looks better on my computer, so it must be a browser specific thing. I'm using Arc (which is Chromium)

erikyo (approved):
LGTM ✈️

src/scripts/components/Drawer.js
erikyo: Mine is Canary Chrome (3 versions ahead of Chrome), maybe too far ahead! 😅

PR #278, opened 2023-04-16
build: add eslint import order rules

johnhooks:
## What?

Add ESLint rules to organize the order of imports in JavaScript files.

## Why?

It helps keep things organized and easier to see at a glance if a decency is from WordPress, external or internal.


erikyo (approved):
LGTM!

PR #279, opened 2023-04-16
fix: eslint config for tests directory

johnhooks:
## What?

Fix the ESLint config for tests directory.

## Why?

The previous config of ESLint didn't include the test directories.


erikyo (dismissed):
LGTM!

erikyo (approved):
LGTM!

PR #283, opened 2023-04-19
fix: rest api namespace

### LINKED ISSUE #194: Change rest-api namespace to `wp-notifications/v1`

johnhooks:
## What?

Move the REST API namespace off the one reserved for WordPress.

## Why?

Until this is merged into core, it should use it's own namespace.

Fixes #194

### DIFF
--- a/includes/restapi/class-notification-controller.php
+++ b/includes/restapi/class-notification-controller.php
@@ -23,7 +23,7 @@ class Notification_Controller extends WP_REST_Controller {
 	 *
 	 * @type string
 	 */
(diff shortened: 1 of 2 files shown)


erikyo (approved):
LGTM!

PR #259, opened 2023-04-12
feature: mvp

johnhooks:
## What?

Create a minimum viable product of the notification plugin.

## Why?

There is still a lot to learn, but we need a minimally functional system to explore the possiblities.

## How?

Combines PRs:

- #178 
- #298 
- #204 
- #214  
- #291
- #255 
- #256 

## Necessary functionality for MVP*

- [x] Activate
- [x] Uninstall
- [ ] Classes
   - [x] Channel
   - [x] Channel_Registry
   - [x] Message
   - [x] Message_Factory
   - [x] Notice
   - [x] Notice_Factory
   - [x] Subscription 
   - [x] Subscription_Factory
- [ ] Database
  - [x] Messages table
  - [x] Queue table
  - [x] Subscriptions table
  - [x] Message_Repository
  - [ ] Notice_Repository (the queue table)
  - [x] Subscription_Repository
- [ ] REST API
  - [ ] `Notification_Controller` `wp-notifications/v1/notifications`
    - [x] `GET /notifications/`
    - [ ] `GET /notifications/{id}`
    - [ ] `POST /notifications/{id}`
    - [ ] `PUT /notifications /{id}`
    - [ ] `DELETE  /notifications /{id}`
  - [ ] `Subscription_Controller` `wp-notifications/v1/subscriptions` 
    - [ ] `GET /subscriptions/`
    - [ ] `GET /subscriptions/{id}`
    - [ ] `POST /subscriptions/{id}`
    - [ ] `PUT /subscriptions /{id}`
    - [ ] `DELETE  /subscriptions /{id}`
- [ ] CLI Command
  - [x] Seed users
  - [x] Seed subscriptions
  - [ ] Seed notification messages
  - [x] List registered channels 
  - [ ] List user's notifications
  - [ ] List user's subscriptions

* Only a rough outline right now, and heavily concentrated on the backend.

PR #289, opened 2023-04-20
feature(mvp): Stage 2 - add initial model classes

johnhooks:
## What?

Add the initial models

- `Model\Channel`
- `Model\Message`
- `Model\Notification`
- `Model\Subscription`

## Why?

These are the base of models developed in #259. 

## Notes

All the properties are nullable, because I intend to support the `_fields` rest option, which means the models might not be initialized with all of their properties, just what the api has requested. I’ll build that in later, but currently the `jsonSerialize` function always returns all property’s of the model, I think this is fine for now. I will build out a validation system that will actually ensure the model properties are valid on insert and update.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -2,6 +2,9 @@
 	"name": "wordpress/wp-feature-notifications",
 	"description": "Notifications for WordPress (Feature Plugin)",
 	"type": "wordpress-plugin",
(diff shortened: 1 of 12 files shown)


erikyo (approved):
LGTM 🚀

PR #290, opened 2023-04-20
feature(mvp): Stage 3 - add initial factory classes

johnhooks:
## What?

Add the initial factory classes.

- `Factory\Message`
- `Factory\Notification`
- `Factory\Subscription`
- `Framework\Factory`

## Why?

Builds on the models of #289. Abstracts the creation of model objects using an array argument through the factory.

~Blocked until #289 is merged.~

**Additionally**

- Modified the base `TestCase` to be based on `PHPUnit_Adapter_TestCase`.

### DIFF
--- a/includes/factory/class-message.php
+++ b/includes/factory/class-message.php
@@ -1,0 +1,85 @@
+<?php
+/**
+ * Notifications API:Messages factory class
+ *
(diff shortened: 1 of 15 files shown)


erikyo (approved):
LGTM 🚀

PR #292, opened 2023-04-20
Style: custom props names prefix replaced with "--wp-notifications-*"

erikyo:
## What?
Fixes the naming convention for css custom props

## How?
Replaces all the occurrences of "--wp-notify-*" with "--wp-notifications-*" in custom props names

relative to #197

### DIFF
--- a/src/styles/dashboard/notice.scss
+++ b/src/styles/dashboard/notice.scss
@@ -8,9 +8,9 @@
 	}
 
 	.wp-notification {
-		--wp-notify--hub-spacing-top: clamp(1em, 3vh, 2em);
-		--wp-notify--hub-element-bottom: clamp(1em, 3vh, 2em);
-		--wp-notify--hub-spacing-gap: clamp(2em, 3vh, 3em);
+		--wp-notifications--hub-spacing-top: clamp(1em, 3vh, 2em);
+		--wp-notifications--hub-element-bottom: clamp(1em, 3vh, 2em);
+		--wp-notifications--hub-spacing-gap: clamp(2em, 3vh, 3em);
 
 		width: 100%;
 		box-sizing: border-box;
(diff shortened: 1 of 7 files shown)


src/styles/dashboard/notice.scss
```diff
@@ -8,9 +8,9 @@
 	}
 
 	.wp-notification {
```
johnhooks, line 10: should this classname be `wp-notifications`?

src/styles/hub/notice.scss
```diff
@@ -5,17 +5,17 @@
 
 	.wp-notification {
```
johnhooks, line 6: Same as the previous comment

johnhooks (changes requested):
What is the style of naming CSS classes? Could we use some standard, like BEM?

src/styles/dashboard/notice.scss
erikyo: no because it refer to a single notification... maybe better to rename into `.notification` at this point?

johnhooks: The right way in this whole prefix scheme would be:

```css
.wp-notifications__notification
```

Its not quite BEM, because we are throwing a namespace in there too. it makes sense to me it is a notification element in the `wp-notifications` css namespace.

johnhooks: I looking into how `@wordpress/components` are namespaced and they are just under `.components` maybe we can just use `.notifications`?

erikyo: HMMM 🤔 looks like a little rough to use something so generic! 

Why not jail all the plugin code with `#id + .className`? This way we have a style that always takes priority while remaining fairly compact and readable (ps I hate BEM)

eg.
```
#wp-notifications { 
  .notification {}
}
```

johnhooks: Okay not BEM, but what is the convention you want to use? I think we need something to make it easy to relate the scss to the elements, or at least do something like `@wordpress/components` so the CSS is located close to the component files?

johnhooks: https://github.com/WordPress/gutenberg/tree/73e6403789b278b48dafadbdc0bba737baae7afc/packages/components/src/button

johnhooks: I don't like the idea of namespacing under an ID. otherwise we have to always use that selector when querying for styles within the source with JS.

johnhooks: So let's table this discussion. It will take more thinking.

johnhooks (approved):
LGTM

We will follow up with a discussion about how we should organize the css.

johnhooks (approved):

PR #293, opened 2023-04-20
Resolve trunk history

Sephsekla:
## What?

Nothing.

## Why?

I would like to publish releases from the Trunk branch, however somewhere in the history of the project we've ended up with unrelated histories, which is preventing a PR from Develop into Trunk.

## How?

There are a few different options to fix this, but for now I've gone with the least destructive option, which is to merge Trunk's commits into Develop (even though they don't actually add anything) with `--allow-unrelated-histories`. Probably needs to not be squashed and merged in this case.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,0 +1,1 @@
+# wp-notify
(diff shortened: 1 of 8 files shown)


johnhooks (approved):
LETM (Looks Empty To Me)

johnhooks (approved):

PR #295, opened 2023-04-20
fix: clean the word notify

johnhooks:
## What?

Clean the last remains of the word notify. There are only a few instances of it left, I didn't change it because it was either linking to an external file, the other was in the `storybook/stories/docs/internal-api.md` which didn't make sense to update until we are ready to publish the actual new api.

## Why?

In preparation for 0.2.0 release, we should have this all cleaned up.


src/scripts/demo/index.js
```diff
@@ -49,10 +49,10 @@ window.addEventListener( 'load', () => {
 	 *  Adding an event listener to the form with the id of `wp-notification-metabox-form` that handles "clear all notifications"
 	 */
 	const wpNotificationClearAll = document.getElementById(
-		'clear-all-wp-notify'
+		'clear-all-wp-notifications'
```
Sephsekla, line 52: 👍 good spot

Sephsekla (dismissed):

src/scripts/demo/index.js
johnhooks: sorry @Sephsekla I didn't have it updated... can you do it again?

johnhooks: Also `Cmd+Shift+F` is pretty good on VSCode to find all instances of a word... would never have been able to do it alone.

Sephsekla (dismissed):

erikyo (approved):
LGTMAW! (Looks Good To Me As Well!)

PR #298, opened 2023-04-21
feature(mvp): Stage 4 - add activator and uninstaller

johnhooks:
## What?

Adds plugin activator and uninstaller. The database SQL to create the database tables for the plugin are in the `Database\Schema` static class.

## Why?

Need to be able to install the plugin and have a database to contain the notification data.

Reopening of #199


erikyo (approved):
LGTM!

PR #300, opened 2023-04-22
feature(mvp): add initial channel and subscription rest controllers

johnhooks:
## What?

Add an initial channel rest controller.

## Why?

To allow fetching a list of all available notification channels in the client.

**Additionally**

- Reorganized the initialization of the rest controllers.
- Added tests for the `Channel_Controller` schema.
- Added a missing `icon` schema property to the `Notification_Controller`


erikyo (approved):
LGTM!

PR #308, opened 2023-04-25
refactor: remove the v0.2.0 demo

johnhooks:
Closes #304

## What?

Remove the v0.2.0 demo code

## Why?

It doesn't belong in the MVP.

## How?

Moved initialization code to `includes/load.php`

PR #309, opened 2023-04-25
fix: storybook fake api data

johnhooks:
## What?

Add the demo fake data back in for Storybook

## Why?

It was removed in #308 but was still required for the Storybook.


erikyo (approved):
LGTM!

PR #310, opened 2023-04-26
refactor: initially render bell icon server side

johnhooks:
## What?

Render the bell icon on the server.

## Why?

The bell icon would appear after the JavaScript loaded and cause the icon to jump on every page transition. 

## How?

@erikyo helped me with the code required to render the bell in the admin bar.

## TODO

- [x] The bell icon initially renders a little high and adjusts once the React component is rendered. Some style that is applied on the client isn't in the PHP code.

### DIFF
--- a/src/scripts/components/NotificationHub.js
+++ b/src/scripts/components/NotificationHub.js
@@ -63,8 +63,8 @@ export const NotificationHub = ( { initialActive = false } ) => {
 	}, [ isActive ] );
 
 	return (
-		<div className="ab-item ab-empty-item" tabIndex={ 0 }>
-			<ShortcutProvider
+		<ShortcutProvider className="ab-item ab-empty-item" tabIndex={ 0 }>
+			<div
 				className={ classNames( [
 					'notifications',
 					isActive ? 'active' : '',
@@ -79,7 +79,7 @@ export const NotificationHub = ( { initialActive = false } ) => {
 					focus={ () => setIsActive( true ) }
 					blur={ () => setIsActive( false ) }
 				/>
-			</ShortcutProvider>
-		</div>
+			</div>
+		</ShortcutProvider>
 	);
 };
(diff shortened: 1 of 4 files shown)


src/scripts/components/NotificationHub.js
```diff
@@ -57,21 +63,23 @@ export const NotificationHub = ( { initialActive = false } ) => {
...
+					toggle={ toggleDrawer }
+					isActive={ isActive }
+				/>
+				<Drawer
+					instance={ drawerRef }
+					focus={ () => setIsActive( true ) }
+					blur={ () => setIsActive( false ) }
+				/>
+			</ShortcutProvider>
+		</div>
```
johnhooks: Wrapping this in `div.ab-item.ab-empty-item`  creates the same HTML structure as WordPress creates for the menu item.

src/styles/hub/admin-bar.scss
```diff
@@ -18,18 +18,17 @@
 	}
 
 	/* the bell icon */
-	.ab-item {
+	.hub-icon {
```
johnhooks, line 23: Changing this name to reflect it's purpose, and the case `ab-item` has been moved up higher in the DOM structure.

src/styles/hub/admin-bar.scss
```diff
@@ -18,18 +18,17 @@
...
 		appearance: none;
 		height: 32px;
 		display: block;
-		padding: 0 10px;
```
johnhooks, line 25: The `ab-empty-item` class provides the padding. The `ab-empty-item` class is applied to menu items that [do not have a link](https://github.com/WordPress/wordpress-develop/blob/6d4fc50edc0f0a57b53577d421dcaca5d943a227/src/wp-includes/class-wp-admin-bar.php#L563-L569).

erikyo (approved):
LGTM!

PR #397, opened 2023-08-09
Feature: Notification polling on the client

### LINKED ISSUE #306: Implement notification polling on the client
### What problem does this address?

Currently notifications are only fetched on the initial page load, new notifications are not displayed until the next page transition.

### What is your proposed solution?

Poll the REST API for new notifications on set interval.

### Questions

- A what interval should we poll for new notifications?

Yes

filipac:
## What?
This PR adds notification polling, currently based on the heartbeat API from WordPress.

## Why?
This attempts to solve the following issue: https://github.com/WordPress/wp-feature-notifications/issues/306

## How?
This PR adds a new Poller class that abstracts how/when/how often the notifications are polled.
It makes sense to use the existing Heartbeat API from WordPress since it is reliable. The Heartbeat API is a simple server polling API built into WordPress, allowing near-real-time frontend updates. 
[Heartbeat API Docs](https://developer.wordpress.org/plugins/javascript/heartbeat-api/)

The Poller class hooks into the `heartbeat.tick` event, which is triggered on a successful heartbeat (in case the user session expired or there is another error, we also skip useless polls to the API).

There is a parameter that is passed to the Poller class named `shouldSkipFirstBeat` which solves the following edge-case: we load all notifications on page load, but also on some pages, the heartbeat API runs very early on the page load (eg: on post edit/create screen, on post list).
To avoid double, repeated polls to fetch notifications, the first `Poller.SKIP_FIRST_INTERVAL` milliseconds, any heartbeat tick events are ignored.

The Poller instance is added to the `notifications` (`window.wp.notifications`) object, so you can easily do:
 - `window.wp.notifications.poller.stop()` - Stop polling
 - `window.wp.notifications.poller.start()` - Start/resume polling

## Testing Instructions
1. Visit any page on `wp-admin`
2. Wait for a heartbeat tick event
3. Notice the polling request.
4. Visit the posts page. Check that only one call is made on page load.

I am open to suggestions and ideas, also if you want to use a setInterval/setTimeout instead of the Heartbeats API, let pe know. I think it is appropriate though.

Closes #306

### DIFF
--- a/src/poller.ts
+++ b/src/poller.ts
@@ -1,0 +1,46 @@
+export class Poller {
+	/**
+	 * On pages like the post editor or post list, the heartbeat is fired right at the beginning
+	 * of the page load. We already fetched the notifications on page load, so this would result
+	 * in a duplicate request. To prevent this, we skip the first heartbeat tick for the first X seconds.
+	 */
+	SKIP_FIRST_INTERVAL = 5000;
+	public skipFirstBeat = true;
+
+	constructor( shouldSkipFirstBeat = true ) {
+		this.skipFirstBeat = shouldSkipFirstBeat;
+	}
+
+	public pollUpdates() {
+		if ( this.skipFirstBeat ) {
+			return;
+		}
+		window.wp.notifications.fetchUpdates( true );
+	}
+
+	public unsetSkipFirstBeat() {
+		setTimeout( () => {
+			this.skipFirstBeat = false;
+		}, this.SKIP_FIRST_INTERVAL );
+	}
+
+	public start() {
+		if ( this.skipFirstBeat ) {
+			this.unsetSkipFirstBeat();
+		}
+
+		wp.hooks.addAction(
+			'heartbeat.tick',
+			'wp-notifications/pollUpdates',
+			this.pollUpdates.bind( this )
+		);
+	}
+
+	public stop() {
+		wp.hooks.removeAction(
+			'heartbeat.tick',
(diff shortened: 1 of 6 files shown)


Sephsekla:
@filipac my apologies for leaving this unreviewed for so long! Unfortunately things have been on hold a little bit due to team availability and illness, but I'm eager to get things rolling again so will give this a look shortly.

Sephsekla (approved):
This is fantastic work, I can only apologise for how it's taken to look this over properly.

We may well need to iterate on this in future (like a lot of the plugin), but for now I think it's a great approach.

EdithAllison:
Heartbeat may be turned off as it's got a reputation for slowing sites down, and if you google WP performance hacks, disabling or limiting Heartbeat often comes up. 

Plugins that disable or limit Heartbeat (not a comprehensive list):
https://perfmatters.io/
https://wordpress.org/plugins/heartbeat-control/
https://wordpress.org/plugins/sg-cachepress/
https://wordpress.org/plugins/litespeed-cache/
or search on WP plugin directory: https://wordpress.org/plugins/search/heartbeat/

Articles:
https://perfmatters.io/docs/disable-wordpress-heartbeat-api/
https://www.wpbeginner.com/plugins/how-to-limit-heartbeat-api-in-wordpress/
https://www.hostinger.com/tutorials/wordpress-heartbeat

If we rely on Heartbeat we may first need to check it's enabled, and runs often enough. Maybe add a notification / warning if it's turned off.

PR #444, opened 2023-11-15
Update dependabot config to use grouping

Sephsekla:
## What?

Dependabot now supports grouping of PRs based on patterns https://github.blog/changelog/2023-06-30-grouped-version-updates-for-dependabot-public-beta/

This should make update much less tedious and much easier to manage.

## Why?

It's important to keep on top of updates, however we currently have a huge number of pull requests being generated by Dependabot, which are not easy to keep on top of.

## How?

I've added some grouping as follows:

- All `@wordpress/*` npm updates should be grouped
- All phpunit Composer updates should be grouped

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -12,6 +12,11 @@ updates:
           prefix: 'Composer'
       allow:
           - dependency-type: 'direct'
+      groups:
+          phpunit:
+              patterns:
+                  - 'phpunit/phpunit'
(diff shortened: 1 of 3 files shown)


EdithAllison:
@Sephsekla I have never used Dependabot before, struggling to understand the implications hence hesitant to click the approve button but happy to approve if you need me to!

johnhooks (approved):
Looks good to me, group those dependency updates!

PR #479, opened 2023-12-06
Add storybook grouping for Dependabot

Sephsekla:
## What?

Add a new group for Dependabot to group Storybook dependencies

## Why?

There are multiple Storybook packages with updates, this should make this more maintainable and produce less noise.

## How?

Create a new group for npm dependencies under `@storybook/*`

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -36,3 +36,6 @@ updates:
           wordpress:
               patterns:
                   - '@wordpress/*'
+          storybook:
+              patterns:
+                - '@storybook/*'


erikyo:
Hey @Sephsekla, why not setup the renovate bot? https://docs.renovatebot.com/

In this way we can we finally get rid of all those alert in a definitive way!

Sephsekla:
@erikyo OK for me to merge this for now and potentially look at the renovatebot to replace this later?

erikyo (approved):
LGTM! 🚀🎉

erikyo:
> @erikyo OK for me to merge this for now and potentially look at the renovatebot to replace this later?

Yes of course, and sorry for the delay! To me this is obviously fine, however if we can I would use the other bot that does the chores by itself would be cool!

